PR #5, opened 2023-01-19
Incorperate Mjolnir's new Policy List Manager

Gnuxie:
- https://github.com/matrix-org/mjolnir/pull/462
- https://github.com/matrix-org/mjolnir/pull/463

### DIFF
--- a/src/Mjolnir.ts
+++ b/src/Mjolnir.ts
@@ -311,7 +311,7 @@ export class Mjolnir {
             this.protectedRoomsConfig.getExplicitlyProtectedRooms().forEach(this.protectRoom, this);
             // We have to build the policy lists before calling `resyncJoinedRooms` otherwise mjolnir will try to protect
             // every policy list we are already joined to, as mjolnir will not be able to distinguish them from normal rooms.
-            await this.policyListManager.init();
+            await this.policyListManager.start();
             await this.resyncJoinedRooms(false);
             await this.protectionManager.start();
 
--- a/src/Mjolnir.ts
+++ b/src/Mjolnir.ts
@@ -41,7 +41,7 @@ import { WebAPIs } from "./webapis/WebAPIs";
 import RuleServer from "./models/RuleServer";
 import { ThrottlingQueue } from "./queues/ThrottlingQueue";
 import { getDefaultConfig, IConfig } from "./config";
-import { PolicyListManager } from "./models/PolicyList";
+import { PolicyListManager } from "./models/PolicyListManager";
 import { ProtectedRoomsSet } from "./ProtectedRoomsSet";
 import ManagementRoomOutput from "./ManagementRoomOutput";
 import { ProtectionManager } from "./protections/ProtectionManager";
--- a/src/Mjolnir.ts
+++ b/src/Mjolnir.ts
@@ -29,7 +29,6 @@ import {
     LogLevel,
     LogService,
     MembershipEvent,
-    Permalinks,
 } from "matrix-bot-sdk";
 
 import { ALL_RULE_TYPES as ALL_BAN_LIST_RULE_TYPES } from "./models/ListRule";
@@ -42,7 +41,7 @@ import { WebAPIs } from "./webapis/WebAPIs";
 import RuleServer from "./models/RuleServer";
 import { ThrottlingQueue } from "./queues/ThrottlingQueue";
 import { getDefaultConfig, IConfig } from "./config";
-import PolicyList from "./models/PolicyList";
+import { PolicyListManager } from "./models/PolicyList";
 import { ProtectedRoomsSet } from "./ProtectedRoomsSet";
 import ManagementRoomOutput from "./ManagementRoomOutput";
 import { ProtectionManager } from "./protections/ProtectionManager";
@@ -55,9 +54,6 @@ export const STATE_NOT_STARTED = "not_started";
 export const STATE_CHECKING_PERMISSIONS = "checking_permissions";
 export const STATE_SYNCING = "syncing";
 export const STATE_RUNNING = "running";
-
-const WATCHED_LISTS_EVENT_TYPE = "org.matrix.mjolnir.watched_lists";
-const WARN_UNPROTECTED_ROOM_EVENT_PREFIX = "org.matrix.mjolnir.unprotected_room_warning.for.";
 
 /**
  * Synapse will tell us where we last got to on polling reports, so we need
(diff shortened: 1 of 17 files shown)


src/Mjolnir.ts
```diff
@@ -413,7 +406,7 @@ export class Mjolnir {
         // We filter out all policy rooms so that we only protect ones that are
         // explicitly protected, so that we don't try to protect lists that we are just watching.
         const filterOutManagementAndPolicyRooms = (roomId: string) => {
-            const policyListIds = this.policyLists.map(list => list.roomId);
+            const policyListIds = this.policyListManager.lists.map(list => list.roomId);
             return roomId !== this.managementRoomId && !policyListIds.includes(roomId);
         };
```
Gnuxie, line 411: It is a smell that this filter has been ignored and not included as part of the `PolicyListManager`.

src/models/PolicyList.ts
```diff
@@ -607,3 +624,177 @@ export class Revision {
...
+
+            let roomId;
+            try {
+                roomId = await this.mjolnir.client.resolveRoom(permalink.roomIdOrAlias);
+            } catch (ex) {
+                // Let's not fail startup because of a problem resolving a room id or an alias.
+                LogService.warn('Mjolnir', 'Could not resolve policy list room, skipping for this run', permalink.roomIdOrAlias)
+                await this.mjolnir.managementRoomOutput.logMessage(LogLevel.WARN, "Mjolnir", `Room ${permalink.roomIdOrAlias} could **not** be resolved, perhaps a server is down? Skipping this room. If this is a recurring problem, please consider removing this room.`);
+                this.failedStartupWatchListRefs.add(roomRef);
+                continue;
```
Gnuxie, line 742: This needs changing to as per https://github.com/matrix-org/mjolnir/issues/404
We should follow through and add the migration step.

Gnuxie: Do we also need a generic utility for account data and migrating the data format?

test/commands/UnbanBanCommandTest.ts
```diff
@@ -70,7 +71,7 @@ describe("UnbanBanCommand", () => {
 
         it("should be able to detect servers with ban reasons", async () => {
             const mjolnir = createTestMjolnir();
-            (<any>mjolnir).lists = [{listShortcode: "test"}];
+            (<any>mjolnir).policyListManager.lists = [{listShortcode: "test"}];
```
Gnuxie, line 74: While I think there has been a precedent for exposing these utilities externally from Mjolnir, I think we should at least try keeping them internal to Mjolnir and delegating.

src/Mjolnir.ts
Gnuxie: Ok, that's because it's to do with protected rooms, not the policy list manager... maybe it could be moved there though?

src/Mjolnir.ts
```diff
@@ -562,7 +452,7 @@ export class Mjolnir {
 
         // Check for updated ban lists before checking protected rooms - the ban lists might be protected
         // themselves.
-        const policyList = this.policyLists.find(list => list.roomId === roomId);
+        const policyList = this.policyListManager.lists.find(list => list.roomId === roomId);
```
Gnuxie, line 455: The direct access to the lists seem wrong, at the least we should ensure `.lists` returns a shallow copy. But we should think about avoiding access all together.

Also see https://github.com/Gnuxie/Draupnir/pull/5/files#r1086994746 which has a similar access above

src/models/MatrixDataManager.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
Gnuxie, line 1: This file isn't actually used by the new update, i thought that it would be needed (which is why it was created), because we now resolve aliases before storing them.

src/models/PolicyList.ts
```diff
@@ -25,13 +25,32 @@ limitations under the License.
...
+ *   references: string[], // Each entry is a `matrix.to` permalink.
+ * }
+ * ```
+ */
+export const WATCHED_LISTS_EVENT_TYPE = "org.matrix.mjolnir.watched_lists";
+
+/**
+ * A prefix used to record that we have already warned at least once that a PolicyList room is unprotected.
+ */
+const WARN_UNPROTECTED_ROOM_EVENT_PREFIX = "org.matrix.mjolnir.unprotected_room_warning.for.";
```
Gnuxie, line 53: Hm? was this always a prefix??? that doesn't seem right

src/models/PolicyList.ts
```diff
@@ -607,3 +626,160 @@ export class Revision {
         return this.ulid > revision.ulid;
     }
 }
+
+type WatchedListsEvent = RawSchemedData & { references?: string[] };
+
+/**
+ * A manager for all the policy lists for this Mjölnir
```
Gnuxie, line 633: you're going to need to be a bit more descriptive than that

src/Mjolnir.ts
```diff
@@ -444,109 +437,6 @@ export class Mjolnir {
...
-        });
-        return list;
-    }
-
-    public async warnAboutUnprotectedPolicyListRoom(roomId: string) {
-        if (!this.config.protectAllJoinedRooms) return; // doesn't matter
-        if (this.protectedRoomsConfig.getExplicitlyProtectedRooms().includes(roomId)) return; // explicitly protected
-
-        try {
-            const accountData: { warned: boolean } | null = await this.client.getAccountData(WARN_UNPROTECTED_ROOM_EVENT_PREFIX + roomId);
```
Gnuxie, line 508: so apparently it really did always used to work with a prefix huh

src/utils.ts
```diff
@@ -398,16 +398,64 @@ function patchMatrixClientForRetry() {
...
+            }
+
+            if (typeof (response.body) === 'string') {
+                try {
+                    response.body = JSON.stringify(JSON.parse(response.body, jsonReviver));
+                } catch (e) {
+                    // we don't care if we fail to parse the JSON as it probably isn't JSON.
+                }
+            }
+            return cb(error, response, resBody);
```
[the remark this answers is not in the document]
Gnuxie: yes we can

src/models/PolicyList.ts
Gnuxie: yes it was always this way

src/models/PolicyList.ts
Gnuxie: we made one

Gnuxie: `MatrixDataManager`

PR #8, opened 2023-01-25
Incorperate Mjolnir's new report-to-moderator features

Gnuxie:
* https://github.com/matrix-org/mjolnir/pull/465

Mangled in these commits were changes to display name. So that will follow afterwards

Checklist:

- [ ] https://github.com/matrix-org/mjolnir/issues/475
- [ ] https://github.com/matrix-org/mjolnir/issues/471

### DIFF
--- a/mx-tester.yml
+++ b/mx-tester.yml
@@ -75,3 +75,19 @@ homeserver:
     remote:
       per_second: 10000
       burst_count: 10000
+
+# Creating a few users simplifies testing.
+users:
+  - localname: admin
+    admin: true
+    rooms:
+      - public: true
+        name: "List of users"
+        alias: access-control-list
+        members:
+          - admin
+          - user_in_mjolnir_for_all
+  # This user can use Mjölnir-for-all
+  - localname: user_in_mjolnir_for_all
+  # This user cannot
+  - localname: user_regular
--- a/src/commands/SetupDecentralizedReportingCommand.ts
+++ b/src/commands/SetupDecentralizedReportingCommand.ts
@@ -1,0 +1,61 @@
+import { Mjolnir } from "../Mjolnir";
+import { LogLevel } from "matrix-bot-sdk";
+
+const EVENT_MODERATED_BY = "org.matrix.msc3215.room.moderation.moderated_by";
+const EVENT_MODERATOR_OF = "org.matrix.msc3215.room.moderation.moderator_of";
+
+// !mjolnir rooms setup <room alias/ID> reporting
+export async function execSetupProtectedRoom(commandRoomId: string, event: any, mjolnir: Mjolnir, parts: string[]) {
+    // For the moment, we only accept a subcommand `reporting`.
+    if (parts[4] !== 'reporting') {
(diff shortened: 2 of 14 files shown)


Gnuxie:
Related MSC https://github.com/Yoric/matrix-doc/blob/aristotle/proposals/3215-towards-decentralized-moderation.md

Where is the routing bot?

src/commands/SetupDecentralizedReportingCommand.ts
```diff
@@ -0,0 +1,61 @@
+import { Mjolnir } from "../Mjolnir";
+import { LogLevel } from "matrix-bot-sdk";
+
+const EVENT_MODERATED_BY = "org.matrix.msc3215.room.moderation.moderated_by";
+const EVENT_MODERATOR_OF = "org.matrix.msc3215.room.moderation.moderator_of";
+
+// !mjolnir rooms setup <room alias/ID> reporting
+export async function execSetupProtectedRoom(commandRoomId: string, event: any, mjolnir: Mjolnir, parts: string[]) {
```
Gnuxie, line 8: upgrade this to a new style command

mx-tester.yml
```diff
@@ -75,3 +75,19 @@ homeserver:
...
+      - public: true
+        name: "List of users"
+        alias: access-control-list
+        members:
+          - admin
+          - user_in_mjolnir_for_all
+  # This user can use Mjölnir-for-all
+  - localname: user_in_mjolnir_for_all
+  # This user cannot
+  - localname: user_regular
```
Gnuxie, line 93: This probably shouldn't be part of the PR, it's about testing Mjolnir for all

src/protections/LocalAbuseReports.ts
```diff
@@ -0,0 +1,98 @@
...
+                } finally {
+                    // Ignore second exception, propagate first.
+                    throw ex;
+                }
+            }
+        } catch (ex) {
+            mjolnir.managementRoomOutput.logMessage(LogLevel.ERROR, "LocalAbuseReports", ex.message);
+        }
+    }
+}
```
Gnuxie, line 98: newline please

Gnuxie:
So it seems the way this is implemented (which I believe is inaccurate when compared to the MSC) is that the reporter sends an invite to Mjolnir to a DM where the reporter will send the report event.

Unfortunately this just won't work because it conflicts with `config.protectAllJoinedRooms` and allows people to invite the bot account anywhere

Gnuxie:
https://github.com/matrix-org/mjolnir/issues/475

Gnuxie:
Are reports end-to-end encrypted? This is something that makes needing a relay user even more problematic.

I guess for `conig.protectAllJoinedRooms` We can just check who has invited before deciding whether to protect the room, or simply make it incompatible. Alternatively we could ask for a new room type for the relay rooms, but that requires fixing element and the MSC.

Gnuxie:
I say we clone the MSC and have a room type for report tickets. We can consult nheko or fluffy on client UX and develop the event model ourselves with a bot command (or a widget if we're feeling adventurous).

PR #12, opened 2023-02-03
Ban/Unban rework + Prompts for missing arguments

Gnuxie:


src/commands/interface-manager/DeadDocument.ts
```diff
@@ -56,19 +57,24 @@ export enum NodeTag {
...
     UnorderedList = 'ul',
+    OrderedList = 'ol',
     ListItem = 'li',
     LineBreak = 'br',
     BoldFace = 'b',
     ItalicFace = 'i',
+    Anchor = 'a',
 }
 
+// FIXME to implement anchor we need to have a NamedNodeMap
```
Gnuxie, line 68: Fixed, remove this fixme

src/appservice/bot/AppserviceCommandHandler.ts
```diff
@@ -61,7 +61,9 @@ export class AppserviceCommandHandler {
                     appservice: this.appservice,
                     roomId: mxEvent.room_id,
                     event: mxEvent,
-                    client: this.appservice.bridge.getBot().getClient()
+                    client: this.appservice.bridge.getBot().getClient(),
+                    // FIXME: This is WRONG and CANNOT be LEFT.
```
Gnuxie, line 65: Well?

src/commands/Ban.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {list.listShortcode} <a href={list.roomRef}>{list.roomId}</a>
+    </p>
+})
+
+
+export async function findPolicyListFromRoomReference(mjolnir: Mjolnir, policyListReference: MatrixRoomReference): Promise<CommandResult<PolicyList, CommandError>> {
+    // do we need to catch 404 here so we can tell the user the alias was wrong?
+    // Can we shortcut this by making a specific presentaion type for
+    // resolved room id? (that aliases can translate to)
+    const policyListRoomId = (await policyListReference.resolve(this.client)).toRoomIdOrAlias();
```
Gnuxie, line 61: I would catch it and return a command error.

src/commands/Ban.tsx
```diff
@@ -0,0 +1,161 @@
...
+    if (policyList !== undefined) {
+        return CommandResult.Ok(policyList);
+    } else {
+        // Would it be acceptable to create an anonymous policy list that is not being watched
+        // by mjolnir for the purposes of banning / unbanning? unbanning requires loading the rules
+        // but banning doesn't.. so this means you'd want two types of lists
+        // one of which can only be made by a factory which watches them via sync
+        // and makes sure mjolnir is joined.
+        // This refactor would be important for previewing lists regardless.
+        return CommandError.Result(`There is no policy list that Mjolnir is watching for ${policyListReference.toPermalink()}`);
```
Gnuxie, line 72: if the policy list isn't being watched, create an anonymous list. When unbanning do the same but call `list.update` before calling `unban`.

src/commands/interface-manager/JSXFactory.ts
```diff
@@ -4,11 +4,19 @@
...
 
 import { DocumentNode, LeafNode, makeDocumentNode, makeLeafNode, NodeTag, TextNode } from "./DeadDocument";
+import { findPresentationRenderer } from "./DeadDocumentPresentation";
+import { presentationTypeOf } from "./ParamaterParsing";
 
 type rawJSX = DocumentNode|LeafNode|string|number|Array<rawJSX>;
 
 export function JSXFactory(tag: NodeTag, properties: any, ...rawChildren: (DocumentNode|LeafNode|string)[]) {
     const node = makeDocumentNode(tag);
+    // why not just give nodes the properties object??
```
Gnuxie, line 14: Well? `why not just give nodes the properties object??`

src/commands/interface-manager/MatrixPromptForAccept.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <p>
+            No argument was provided for the paramater {paramater.name}, would you like to accept the default?<br/>
+            {defaultPrompt}
+        </p>,
+        this.roomId, this.event, this.client
+    )
+}
+
+// FIXME: <ol> raw tags will not work if the message is sent across events.
+//        If there isn't a start attribute for `ol` then we'll need to take this into our own hands.
```
Gnuxie, line 28: The attribute is called `start`. Not sure how to apply this in the incremental renderer without looking. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attr-start

src/commands/interface-manager/MatrixPromptForAccept.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { BaseFunction, InterfaceCommand } from "./InterfaceCommand";
+import { JSXFactory } from "./JSXFactory";
+import { MatrixContext } from "./MatrixInterfaceAdaptor";
+import { PromptResponseListener } from "./MatrixPromptUX";
+import { ParamaterDescription } from "./ParamaterParsing";
+import { PromptOptions } from "./PromptForAccept";
+import { CommandResult } from "./Validation";
+
+// How are prompts going to work?
+// We need to temporarily hook onto the event emitter.
```
Gnuxie, line 16: Remove this command, we've done it.

src/commands/interface-manager/MatrixPromptUX.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Copyright (C) 2023 Gnuxie <Gnuxie@protonmail.com>
+ * All rights reserved.
+ */
+
+import { MatrixEmitter } from "../../MatrixEmitter";
+import { CommandError, CommandResult } from "./Validation";
+import { LogService } from "matrix-bot-sdk";
+
+// Internal to PromptResponseListener and needs to be manually managed
+// ie calls to .on and .off, so this is why it's internal
```
Gnuxie, line 11: done

src/commands/interface-manager/ParamaterParsing.ts
```diff
@@ -24,15 +24,34 @@ limitations under the License.
...
+        return false;
+    }
+
+    prompt(paramaterDescription: ParamaterDescription): Promise<CommandResult<ReadItem>> {
+        throw new TypeError("This argument stream is NOT promptable, did you even check isPromptable().");
+    }
 }
 
+// TODO: Presentation types should be extracted to their own file.
+// FIXME: PresentationTypes should not be limited to ReadItems.
```
Gnuxie, line 53: Feel like this needs to be a follow up along with presentation type translators

src/commands/interface-manager/MatrixPromptUX.ts
```diff
@@ -0,0 +1,171 @@
...
+        if (index < 10) {
+            return ["1️⃣", "2️⃣", "3️⃣", "4️⃣", "5️⃣", "6️⃣", "7️⃣", "8️⃣", "9️⃣", "🔟"][index];
+        } else {
+            return index.toString();
+        }
+    }
+
+    // This won't work, we have to have a special key in the original event
+    // that means we should be waiting for it, that can't be abused/forged.
+    // As we can't have the event id AOT.
```
Gnuxie, line 161: Another followup, but might be superstitious

src/commands/interface-manager/PromptForAccept.ts
```diff
@@ -0,0 +1,58 @@
...
+        return this.source.slice(this.position);
+    }
+
+    public isPromptable(): boolean {
+        return this.interfaceAcceptor.isPromptable
+    }
+
+    public async prompt<T = ReadItem>(paramaterDescription: ParamaterDescription): Promise<CommandResult<T>> {
+        // FIXME I thought prompt for accept could return multiple items if there were
+        // othere paramaters after this one, but never mind..
```
Gnuxie, line 48: This can only work with replies, i think it's unnecessary, the prompts are pretty quick.

src/commands/interface-manager/MatrixPromptUX.ts
```diff
@@ -0,0 +1,171 @@
...
+        if (!(typeof relatedEventId === 'string' && typeof reactionKey === 'string')) {
+            return;
+        }
+        const entry = this.promptRecordByEvent.get(relatedEventId);
+        if (entry !== undefined) {
+            for (const record of entry) {
+                const presentation = record.presentationByReaction.get(reactionKey);
+                if (presentation === undefined) {
+                    // FIXME: Should this be WARN? Technically the prompt should fail as saying
+                    // that the reaction wasn't understood.
```
Gnuxie, line 85: Hmmmm I feel like i just deleted this comment thinking it was asking for a warn rather than a fail

src/commands/interface-manager/MatrixPromptUX.ts
```diff
@@ -0,0 +1,180 @@
...
+            promptRecords.delete(promptRecord);
+            if (promptRecords.size === 0) {
+                this.promptRecordByEvent.delete(eventId);
+            }
+        }
+    }
+
+    private async addBaseReactionsToEvent(roomId: string, eventId: string, presentationsByKey: PresentationByReactionKey) {
+        return await Promise.all([...presentationsByKey.keys()].slice(0, 5)
+            .map(key => this.client.unstableApis.addReactionToEvent(roomId, eventId, key)));
```
Gnuxie, line 66: Can we ensure ordering without it being dog slow?

src/commands/Ban.tsx
Gnuxie: I would agree, but it is rather inconvenient to do without wrapping in try catch with a let or making a `CommandResult` unnecessarily. Quite annoyed. 

```
    const resolvedReference = await policyListReference.resolve(this.client)
        .catch(e => e.statusCode === 404 ? null : Promise.reject(e));
    if (resolvedReference === null) {
        return CommandError.Result(`Failed to resolve the room alias that was provided`);
    }
```

it is possible to do something like this but meh it doesn't seem right

src/commands/interface-manager/MatrixPromptForAccept.tsx
Gnuxie: Maybe would require follow up, but to do this what you would need is: the fringe walker in `DeadDocument.ts` would have to borrow the dynamic environment from -- the same file, and it already does. So then, it would have to bind the item count when there is an `ol` or at least read it? 

Since i think the markdown renderer and html renderer both rely on this variable. So then you'll probably also have to make a common set of variables for handling ordered lists.

src/commands/interface-manager/ParamaterParsing.ts
Gnuxie: At the least an issue should be created saying that we need to allow the argument stream to contain things other than `ReadItem`s

src/commands/interface-manager/MatrixPromptUX.ts
Gnuxie: Honestly in the grand scheme of things this does not matter

src/commands/interface-manager/MatrixPromptUX.ts
Gnuxie: Yeah just reverse it

PR #20, opened 2023-02-10
Include version information within `!draupnir status`

Gnuxie:
image


Gnuxie:
The tests rely on the first reply to a command to determine the event that is responding to a given command invocation. Since we don't have replies figured out for commands yet, this could be a blocker.

PR #34, opened 2023-02-21
Add pre-commit configuration

Mikaela:
1. Install `pre-commit`, I installed it from Fedora repositories as it's widely packaged.
2. Run `pre-commit` install so the local repository gets the hooks in the right place
3. Run `pre-commit run --all-files` to see it in action
4. When running `git commit`, it runs automatically.

Note!

1. The `.pre-commit-config.yaml` violates the `.editorconfig` and thus doesn't pass its own tests as I had too much trouble indenting to 4 
2. It can be skipped with `git commit --no-verify` which I had to do
3. This wants to touch a lot of files as seen in `git status`
4. as implied in 1. there are 653 errors found
5. This leaves touching CONTRIBUTING instructions a future TODO.

If we hadn't discussed this in Matrix I wouldn't be sure whether this is worth opening a PR on.

Signed-off-by: Aminda Suomalainen <suomalainen+git@mikaela.info>


Mikaela:
Oh, I guess this also makes it possible to enable https://pre-commit.ci/

Mikaela:
https://github.com/Gnuxie/Draupnir/pull/34/commits/1b8e0c27c4549a9c69ebf2adb4e233a39b20c527 drops the amount of errors to 121

Mikaela:
I had ideas and the amount of errors is down to 86.

Mikaela:
The amount of errors is now 0, but I don't know will it run or anything anymore.

```
% pre-commit run --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed
Check .editorconfig rules................................................Passed
```

PR #36, opened 2023-02-22
Ban Propagation protection (that is enabled by default)

Gnuxie:
Inspired by https://github.com/matrix-org/mjolnir/pull/223

image

Later on it should be configurable so that it can automatically add the ban to a list.
In that situation it should also try show the last message of the banned user behind spoiler text so that there is some context for other moderators

### DIFF
--- a/src/commands/interface-manager/DeadDocumentMatrix.ts
+++ b/src/commands/interface-manager/DeadDocumentMatrix.ts
@@ -79,7 +79,7 @@ export async function renderMatrix(node: DocumentNode, cb: SendMatrixEventCB): P
  * @param event An event to reply to.
  * @param client A MatrixClient to send the events with.
  */
-export async function renderMatrixAndSend(node: DocumentNode, roomId: string, event: any, client: MatrixSendClient): Promise<string[]> {
+export async function renderMatrixAndSend(node: DocumentNode, roomId: string, event: any|undefined, client: MatrixSendClient): Promise<string[]> {
     const baseContent = (text: string, html: string) => {
         return {
             msgtype: "m.notice",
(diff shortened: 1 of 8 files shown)


src/commands/interface-manager/DeadDocumentMatrix.ts
```diff
@@ -91,11 +91,14 @@ export async function renderMatrixAndSend(node: DocumentNode, roomId: string, ev
...
-            "m.relates_to": {
-                "m.in_reply_to": {
-                  "event_id": event['event_id']
+            ...event === undefined
+                ? {}
+                : { "m.relates_to": {
+                        "m.in_reply_to": {
+                            "event_id": event['event_id']
+                        }
+                    }
```
Gnuxie, line 100: We should really say why we're doing this or it will be confusing AF

test/integration/commands/commandUtils.ts
```diff
@@ -118,6 +118,31 @@ export async function getFirstReaction(matrix: MatrixEmitter, targetRoom: string
     }
 }
 
+export async function getFirstEventMatching(details: { matrix: MatrixEmitter, targetRoom: string, lookAfterEvent: () => Promise</*event id*/string|undefined>, predicate: (event: any) => boolean }): Promise<any> {
```
Gnuxie, line 121: Please document this

test/integration/commands/commandUtils.ts
```diff
@@ -118,6 +118,31 @@ export async function getFirstReaction(matrix: MatrixEmitter, targetRoom: string
...
+    let targetCb = undefined;
+    try {
+        return await new Promise((resolve, reject) => {
+            details.lookAfterEvent().then((afterEventId: string|undefined) => {
+                let isAfterEventId = afterEventId === undefined;
```
Gnuxie, line 126: This will be confusing af if we don't write why

PR #52, opened 2023-05-10
Update to node 18.

Gnuxie:
End of life was moved forwards by 7 months so we don't have a choice. https://nodejs.org/en/blog/announcements/nodejs16-eol.


Gnuxie:
The problem with CI is that the config says to connect to localhost:8081 which since node 18 prefers to resolve to `::1` rather than `127.0.0.1`. And nginx or something isn't configure to t listen on `::1` but only `127.0.0.1`. https://github.com/nodejs/node/issues/40702

PR #53, opened 2023-05-15
Introduce concept of a synchronization unit that will supersede the management room.

Gnuxie:
The management room will basically become a "synchronization unit". Synchronization units will work in a similar way as policy lists but will determine how different policy lists will be applied to rooms. In a consistent way.

The idea for this is not really solve problems such as "how do i moderate these two spaces differently". More how do we talk about how policy lists are "watched" by rooms and how policies get applied. The idea is to try get to a situation where policies can be applied from lists arbitrarily (with a modular plugin system within Draupnir somewhat like a protection). At the least it opens up room for some immediate UX improvements (by allowing clients to introspect Draupnir directly). 

This PR isn't ready. This isn't the only change going to land on this branch.

### DIFF
--- a/src/models/PolicyList.ts
+++ b/src/models/PolicyList.ts
@@ -33,18 +33,6 @@ import AwaitLock from "await-lock";
 import { monotonicFactory } from "ulidx";
 
 /**
- * Account data event type used to store the permalinks to each of the policylists.
- *
- * Content:
- * ```jsonc
- * {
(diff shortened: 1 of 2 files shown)


Gnuxie:
This work ended up being blocked by an attempt to create a generic library that could be used by both widgets and draupnir.

Gnuxie:
This ended up being [matrix-protection-suite](https://github.com/Gnuxie/matrix-protection-suite)

PR #58, opened 2023-06-12
Add building arm images.

MTRNord:
Based on https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
https://github.com/docker/build-push-action and https://docs.docker.com/engine/reference/commandline/buildx_build/#platform


Gnuxie:
Thanks :medal_sports:

PR #61, opened 2023-06-22
Add a missing dash to the docker command in the appservice documentation

MTRNord:
Without the dash it gives an error.


Gnuxie:
Can you remember if this was really the case?

MTRNord:
At least in newer docker for me `-rm` doesnt exist while `--rm` exists. Not sure if this always has been like this or is a change docker made.

PR #68, opened 2023-07-21
Update to typescript 5

MTRNord:
As discussed elsewhere this brings more speed and also some more nice additions which especially for me downstream are helpful (and later when I upstream more of it)

For others to read: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "mocha": "^9.0.1",
     "ts-mocha": "^9.0.2",
     "tslint": "^6.1.3",
-    "typescript": "^4.9.4",
+    "typescript": "^5.1.6",
     "typescript-formatter": "^7.2"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Gnuxie (approved):
Hello, Looks good, thank you.

Do you know why so many of the packages have changed registry in the resolved package URL from `registry.yarnpkg.com` to `registry.npmjs.org`? Seems weird that there were co many using both before. Not a blocker or anything just don't understand what's going on there.

MTRNord:
> Hello, Looks good, thank you.
> 
> Do you know why so many of the packages have changed registry in the resolved package URL from `registry.yarnpkg.com` to `registry.npmjs.org`? Seems weird that there were co many using both before. Not a blocker or anything just don't understand what's going on there.

Oh likely because I used npm for it. And it wrote the yarn. I can redo this with yarn I guess.

PR #70, opened 2023-07-21
Add health endpoint to appservice and add metrics via prometheus

MTRNord:
This adds a `/healthz` endpoint to the appservice which allows this to work more nicely in kubernetes.

It also adds some metrics for tracking the provisioning state.

Grafana result:
image

Note: The ts-ignore are sadly required since the `_getValue` method is not public :/ I didnt find another solution apart from tracking it maybe elsewhere.

### DIFF
--- a/src/appservice/AppService.ts
+++ b/src/appservice/AppService.ts
@@ -34,6 +34,7 @@ import { IConfig } from "./config/config";
 import { AccessControl } from "./AccessControl";
 import { AppserviceCommandHandler } from "./bot/AppserviceCommandHandler";
 import { SOFTWARE_VERSION } from "../config";
+import { Registry } from 'prom-client';
 
 const log = new Logger("AppService");
 /**
@@ -86,13 +87,18 @@ export class MjolnirAppService {
         const accessControlListId = await bridge.getBot().getClient().resolveRoom(config.adminRoom);
         const accessControl = await AccessControl.setupAccessControl(accessControlListId, bridge);
         // Activate /metrics endpoint for Prometheus
+
+        // This should happen automatically but in testing this didn't happen in the docker image
         setBridgeVersion(SOFTWARE_VERSION);
-        const prometheus = bridge.getPrometheusMetrics(false);
+
+        // Due to the way the tests and this prom library works we need to explicitly create a new one each time.
+        const prometheus = bridge.getPrometheusMetrics(true, new Registry());
         const instanceCountGauge = prometheus.addGauge({
             name: "draupnir_instances",
             help: "Count of Draupnir Instances",
             labels: ["status", "uuid"],
         });
+
         const mjolnirManager = await MjolnirManager.makeMjolnirManager(dataStore, bridge, accessControl, instanceCountGauge);
         const appService = new MjolnirAppService(
             config,
--- a/src/appservice/AppService.ts
+++ b/src/appservice/AppService.ts
@@ -25,7 +25,7 @@ limitations under the License.
  * are NOT distributed, contributed, committed, or licensed under the Apache License.
  */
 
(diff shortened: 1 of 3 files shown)


MTRNord:
Hm it seems like the tests cause the endpoint to be registered more than once :D

> Error: A metric with the name bridge_app_version has already been registered.

Gnuxie:
> Hm it seems like the tests cause the endpoint to be registered more than once :D
> 
> > Error: A metric with the name bridge_app_version has already been registered.

Well, they setup, start and restart the appservice multiple times within the lifetime of the test suite so that's probably way

MTRNord:
Hm maybe a try catch works here.

MTRNord:
Hm it might be a little harder to fix than a try catch since the addGauge seems to be the issue inside of `getPrometheusMetrics`

MTRNord:
@Gnuxie should be all fixed now :)

src/appservice/AppService.ts
```diff
@@ -75,21 +78,35 @@ export class MjolnirAppService {
...
             suppressEcho: false,
         });
         await bridge.initialise();
         const accessControlListId = await bridge.getBot().getClient().resolveRoom(config.adminRoom);
         const accessControl = await AccessControl.setupAccessControl(accessControlListId, bridge);
-        const mjolnirManager = await MjolnirManager.makeMjolnirManager(dataStore, bridge, accessControl);
+        // Activate /metrics endpoint for Prometheus
+
+        // This should happen automatically but in testing this didn't happen in the docker image
+        setBridgeVersion(SOFTWARE_VERSION);
```
Gnuxie, line 92: Are you happy that this can say "A version was either not provided when building Draupnir or could not be read." in an error state? If so that's cool, just checking you knew that's what it does.

src/appservice/MjolnirManager.ts
```diff
@@ -70,6 +72,18 @@ export class MjolnirManager {
...
+            this.instanceCountGauge.dec({ status: "offline", uuid: localPart });
+        }
+        // @ts-ignore
+        if (this.instanceCountGauge._getValue({ status: "disabled", uuid: localPart })) {
+            this.instanceCountGauge.dec({ status: "disabled", uuid: localPart });
+        }
+        // @ts-ignore
+        if (!this.instanceCountGauge._getValue({ status: "online", uuid: localPart })) {
+            this.instanceCountGauge.inc({ status: "online", uuid: localPart });
+        }
```
Gnuxie, line 86: Not that we can't accommodate for this, but what is going on here?

MTRNord: so the `_getValue` is private in the gauge. One way would be to cache it on the draupnir side again. But it seems rather weird to have 2 variables saving the same state here to me. So I went with evil mode here.

I could see why it might make sense to rewrite it though. (Tbh its a little weird to me why getValue is private in the upstream dependency. But its not like we can change that easily)

src/appservice/AppService.ts
MTRNord: I think its fine. In the end its just a string thats mandatory by the bridge when using the metrics stuff. It actually is only called because for some reason the path it checks in the bridge sdk is not pointing to the package.json when draupnir is in docker. (It does work when outside of docker though)

src/appservice/MjolnirManager.ts
Gnuxie: Ok, that's ok. If it makes sense to you, i'd probably localise the private access to something like an instanceCountGuage to a function/method, e.g. `getGuageValue`, keeping the ts-ignore within that function and then add some context in comments for why it's necessary. What do you think?

MTRNord: Sounds like a good idea. I will have a look at that on the weekend

Gnuxie (approved):
LGTM :rocket:

Issue #83, opened 2023-08-30
Ban propagation protection should also prompt for unbans when there are rules matching the user.


Gnuxie:
Makes sense that someone struggling with unbanning a user would try to unban them from a room without checking the rules. It used to happen all the time with Mjolnir because of the confusing `--force` requirement of the unban command.

Gnuxie:
If there was another list, where the user is banned that you don't have permission to remove the rule, you'd have to give up anyway or unwatch that list.

PR #85, opened 2023-08-31
Revisit Mjolnir's ErrorCache.

Gnuxie:
Read the commits. Not finished yet.

- [x] We probably need to include a patch for the toggle to disable server ACL within the same PR.

- [x] We needed both the removal of the error cache and the changes to `printActionResult` for an attempt at #83 . Since it's something that we'd want to reuse and comfort users with as a common pattern when they are "changing something" . I guess we're also in a position to consider changing what `printActionResult` actually shows, now that it is possible to change it consistently and safely. 

Actually another reason why the ErrorCache was added was because if Mjolnir had been offline and is unable to ban a bunch of new members when it comes back up (receiving lots of join events from sync) it will spam the exact errors for every member event). Maybe the only way to fix that is to add some kind of queue like the acl batching has. 
It also gets the errors because Draupnir forcibly verifies permissions at startup and performs a full sync. There isn't an easy way to manage this. Maybe bring back the error cache but make it more specific ? Include variants for unban / ban / whatever

the problem really is "verify that these errors are the same errors as before" 
i'd like to think it wouldn't be a problem if i could do something like "multiple errors are happening see https://eijfoj/" and that page can be overwritten etc. But it would still be a problem e.g. if you wanted to ping someone when there are errors, because you'd have to verify that the same errors were happening to decide whether they need to be pinged again. 
We could be lucky and this is a problem people have to solve all the time and is well documented.


src/models/RoomUpdateError.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * @param client A matrix client.
+ * @param errors Any errors associated with the action.
+ * @param options.title To give context about what the action was, shown when there are errors.
+ * @param options.noErrorsText To show when there are no errors.
+ * @returns A `DocumentNode` fragment that can be sent to Matrix or incorperated into another message.
+ */
+export async function renderActionResult(
+    client: MatrixSendClient,
+    errors: RoomUpdateError[],
+    { title = '', noErrorsText = 'Done updating rooms - no errors.'}: { title?: string, noErrorsText?: string }
```
Gnuxie, line 60:
```suggestion
    { title = '', noErrorsText = 'Done updating rooms - no errors.'} = {}: { title?: string, noErrorsText?: string }
```

src/ProtectedRoomsSet.ts
```diff
@@ -205,51 +202,33 @@ export class ProtectedRoomsSet {
...
             // Only apply bans and then redactions in the room we are currently looking at.
-            const banErrors = await this.applyUserBans([roomId]);
-            const redactionErrors = await this.processRedactionQueue(roomId);
-            await this.printActionResult(banErrors);
-            await this.printActionResult(redactionErrors);
+            const errors = [
+                await this.applyUserBans([roomId]),
+                await this.processRedactionQueue(roomId),
+            ].flat();
+            await this.printActionResult(errors, { title: 'Problem updating member bans.' });
```
Gnuxie, line 218: `printActionResult` didn't used to display a message when there were no errors so this might be a problem here

config/default.yaml
```diff
@@ -73,6 +73,10 @@ verifyPermissionsOnStartup: true
 # turn on to trial some untrusted configuration or lists.
 noop: false
 
+# Whether or not Draupnir should apply `m.room.server_acl` events.
+# DO NOT change this to `true` unless you are very confident that you know what you are doing.
+disableServerACL: false
```
Gnuxie, line 78: Reminder: need to change verify permissions too

Gnuxie:
I'm going to merge because to me it's already pretty bad if there are persistent errors. And it'd be better to do work to ensure that there aren't.

PR #87, opened 2023-09-01
Add eslint to pre-commit

Mikaela:


.pre-commit-config.yaml
```diff
@@ -9,6 +9,18 @@ repos:
       - id: end-of-file-fixer
       - id: check-yaml
       - id: check-added-large-files
+  - repo: https://github.com/pre-commit/mirrors-eslint
+    rev: "v8.48.0"
+    hooks:
+      - id: eslint
+        files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
+        types: [file]
```
Mikaela, line 17: This comes from upstream documentation

- https://github.com/pre-commit/mirrors-eslint#using-eslint-with-pre-commit

although I think just `files: [ts]` should be enough as `identify-cli` recognises ts files as such.

MTRNord: Draupnir has `ts` and `tsx` file endings :) So at least `\.tsx?$` is needed

Gnuxie:
Thanks Aminda

PR #88, opened 2023-09-01
Move to eslint

MTRNord:
This moves the linter to ESLint. The left over tslint.json is intentional since we then can reuse it to generate the prettier config later from it. After that, it can be removed.

For converting, `npx tslint-to-eslint-config --comments tsconfig.json` was used.

To quote it's output:

```
3 ESLint rules behave differently from their TSLint counterparts:
  * @typescript-eslint/no-unused-expressions:
    - The TSLint optional config "allow-new" is the default ESLint behavior and will no longer be ignored.
  * brace-style:
    - ESLint's brace-style will check all tokens.
  * no-underscore-dangle:
    - Leading or trailing underscores (_) on identifiers will now be forbidden.

1 rule is not known by tslint-to-eslint-config to have an ESLint equivalent:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "whitespace".
```


Gnuxie:
I'll merge this after https://github.com/Gnuxie/Draupnir/pull/85 and unban propagation rework. I'm really worried about somehow weakening the requirements accidentally.

PR #89, opened 2023-09-01
Some safety around ACL allows.

Gnuxie:
Related: https://github.com/Gnuxie/Draupnir/issues/56.

Needs some kind of testing before merging.

### DIFF
--- a/src/models/ServerAcl.ts
+++ b/src/models/ServerAcl.ts
@@ -26,7 +26,6 @@ limitations under the License.
  */
 
 import { MatrixGlob } from "matrix-bot-sdk";
-import { setToArray } from "../utils";
 
 export interface ServerAclContent {
     allow: string[];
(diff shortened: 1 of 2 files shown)


src/models/ServerAcl.ts
```diff
@@ -104,6 +104,9 @@ export class ServerAcl {
         if (!allowed || allowed.length === 0) {
```
Gnuxie, line 104: Wait a minute, `!allowed`?? What's that supposed to mean? And how are we pushing to it if its undefined??

TheArcaneBrony: Probably should make this `allowed = [ '*' ];`

Gnuxie: Why?

As for why `!allowed` was there, seemingly it was a redundant check from a long time ago

Gnuxie:
This was incorporated into MPS

Issue #92, opened 2023-09-04
The legacy of `ErrorCache`: What if there are situations where errors persist?


Gnuxie:
A remaining unmitigated reason for why ErrorCache was added was because if Mjolnir had been offline and is unable to ban a bunch of new members when it comes back up (receiving lots of join events from sync) it will spam the exact errors for every member event). The management room also reports the errors because Draupnir forcibly verifies permissions at startup and performs a full sync.

This problem really is "find some way to classify whether these errors are the same errors sorts of errors as before, ie is it the same problem and do i already know about it?"
We could be lucky and this is a problem people have to solve all the time and is well documented.

Follow up from https://github.com/Gnuxie/Draupnir/pull/85

MTRNord:
My mind drifted to "throw it into a set and forget" 😅 since a set would just drop it if it already exists. Not sure if that's a good solution though

Gnuxie:
I don't think we really have this issue in v2.0.0? We're much smarter about when we run protections.

PR #93, opened 2023-09-04
Unban propagation

Gnuxie:
Fixes https://github.com/Gnuxie/Draupnir/issues/83
Fixes https://github.com/Gnuxie/Draupnir/issues/84
Related #92 

Extends the ban propagation to offer to unban a user from all protected rooms when an unban is detected in a protected room.

image

Thoughts:

- [x] The pattern for using `printActionResult` is common but lacks context about what caused `RoomUpdateErrors`, including whether they are expected or unexpected. We also don't utilise `CommandError/Result` which would give the id for admins to track the error if it was unexpected.
- [x] It's not clear which situations we should use the log level `ERROR` or `INFO` depending on what caused unban to fail (similar to CommandError, which has a distinction between reporting known errors and unexpected errors)
-  Collecting RoomUpdateErrors is a common pattern.
- There's probably a lot of duplicated effort between the ban and unban listeners

### DIFF
--- a/src/ProtectedRoomsSet.ts
+++ b/src/ProtectedRoomsSet.ts
@@ -25,14 +25,15 @@ limitations under the License.
  * are NOT distributed, contributed, committed, or licensed under the Apache License.
  */
 
-import { LogLevel, LogService, MatrixGlob, UserID } from "matrix-bot-sdk";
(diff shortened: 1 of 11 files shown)


Gnuxie:
It looks like a big PR but quite a lot of it comes from boiler plate in the protection itself

PR #103, opened 2023-09-09
.github/dependabot.yml: explicitly configure dependabot

Mikaela:
@MTRNord says that there is a warning in logs about missing dependabot.yml.

I know that this says `npm`, but https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file says that it also does yarn.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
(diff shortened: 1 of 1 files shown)


MTRNord:
Well. To be exact there would be a warning if activated. But apparently the move disabled it anyway. So it doesn't really matter what those logs would have told us :)

Gnuxie (approved):

Gnuxie:
Just to be clear, the reason this is necessary is because moving the repository seemed to have disabled the alerts

Issue #113, opened 2023-09-11
Dependabot: Group dev dependencies and dependencies for github actions


Gnuxie:
Should make managing a bit easier 

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/

PR #114, opened 2023-09-11
.github/dependabot.yml: attempt to group updates

Mikaela:
Resolves: #113


Gnuxie:
It can't get any worse can it?

PR #129, opened 2023-09-13
Bump the dockerfile to node 20

MTRNord:


Gnuxie:
The CI is supposed to use the lowest supported version (like the readme and package JSON) so it's probably best to leave that at 18. Dockerfile is fine to be 20 though.

MTRNord:
> The CI is supposed to use the lowest supported version (like the readme and package JSON) so it's probably best to leave that at 18. Dockerfile is fine to be 20 though.

oh. makes sense

PR #144, opened 2023-10-14
Try to reduce usage of `any` in the codebase

MTRNord:
This also makes the usage of `any` be a linter error by default. In most cases, we likely don't want `any`. Most usages that currently exist seem to be from `matrix-bot-sdk`. Some are from nodejs itself.

### DIFF
--- a/src/commands/SinceCommand.ts
+++ b/src/commands/SinceCommand.ts
@@ -213,7 +213,8 @@ async function execSinceCommandAux(destinationRoomId: string, event: any, mjolni
             } else if (maybeRoom.startsWith("#") || maybeRoom.startsWith("!")) {
                 const roomId = await mjolnir.client.resolveRoom(maybeRoom);
                 if (!protectedRooms.has(roomId)) {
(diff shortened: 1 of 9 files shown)


src/commands/SinceCommand.ts
```diff
@@ -213,7 +213,8 @@ async function execSinceCommandAux(destinationRoomId: string, event: any, mjolni
             } else if (maybeRoom.startsWith("#") || maybeRoom.startsWith("!")) {
                 const roomId = await mjolnir.client.resolveRoom(maybeRoom);
                 if (!protectedRooms.has(roomId)) {
-                    return mjolnir.managementRoomOutput.logMessage(LogLevel.WARN, "SinceCommand", `This room is not protected: ${htmlEscape(roomId)}.`);
```
MTRNord, line 216: This was found as a result of the any return type of logMessage. I fixed it with what I assume may have been intended here. But logMessage can only return void. So the return made no selve (and now causes a typescript error as it should)

Gnuxie:
I don't think this fundamentally changes anything. I know that having to manually disable the check would stop `any` creeping into new code and urge caution, but I don't know if this is the right move yet. Ultimately I want to be able to parse all incoming events so that we don't have this problem but that's probably not going to happen anytime soon. For now though encouraging `unknown` over any would be nice

PR #145, opened 2023-10-14
Add support for bot-sdk crypto as an experimental flag

MTRNord:
Adds `experimentalRustCrypto` as a boolean flag to the config. As this obviously is mutually exclusive with pantalaimon it does disable pantalaimon if used. It also issues a warning if `config.pantalaimon.use` is also true. Since that likely is invalid, config. However, as long as `accessToken` is also set, it will work. Otherwise, it fails regardless trying to log in.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -28,6 +28,13 @@ pantalaimon:
   # After successfully logging in once, this will be ignored, so this value can be blanked after first startup.
   password: your_password
 
+# Experimental usage of the matrix-bot-sdk rust crypto.
+# This can not be used with Pantalaimon.
+# Make sure to setup the bot as if you are not using pantalaimon for this.
+#
+# Warning: At this time this is not considered production safe.
+experimentalRustCrypto: false
+
 # The path Draupnir will store its state/data in, leave default ("/data/storage") when using containers.
 dataPath: "/data/storage"
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -129,6 +129,9 @@ export interface IConfig {
             enabled: boolean;
         }
     }
+    // Experimental usage of the matrix-bot-sdk rust crypto.
+    // This can not be used with Pantalaimon.
+    experimentalRustCrypto: boolean;
 
     /**
      * Config options only set at runtime. Try to avoid using the objects
@@ -210,6 +213,7 @@ const defaultConfig: IConfig = {
             enabled: false,
         },
     },
+    experimentalRustCrypto: false,
 
     // Needed to make the interface happy.
     RUNTIME: {
@@ -225,7 +229,7 @@ export function getDefaultConfig(): IConfig {
  * @param argv An arguments vector sourced from `process.argv`.
  * @returns The path if one was provided or undefined.
  */
-function configPathFromArguments(argv: string[]): undefined|string {
+function configPathFromArguments(argv: string[]): undefined | string {
     const configOptionIndex = argv.findIndex(arg => arg === "--draupnir-config");
     if (configOptionIndex > 0) {
         const configOptionPath = argv.at(configOptionIndex + 1);
--- a/src/index.ts
+++ b/src/index.ts
@@ -35,7 +35,8 @@ import {
     MatrixClient,
     PantalaimonClient,
     RichConsoleLogger,
-    SimpleFsStorageProvider
+    SimpleFsStorageProvider,
+    RustSdkCryptoStorageProvider
 } from "matrix-bot-sdk";
 
 import { read as configRead } from "./config";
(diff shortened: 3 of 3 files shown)


src/index.ts
```diff
@@ -35,7 +35,8 @@ import {
     MatrixClient,
     PantalaimonClient,
     RichConsoleLogger,
-    SimpleFsStorageProvider
+    SimpleFsStorageProvider,
+    RustSdkCryptoStorageProvider
 } from "matrix-bot-sdk";
```
Gnuxie, line 41:
```suggestion
import { StoreType } from "@matrix-org/matrix-sdk-crypto-nodejs";
```

src/index.ts
```diff
@@ -69,9 +70,17 @@ import { initializeSentry, patchMatrixClient } from "./utils";
...
-        if (config.pantalaimon.use) {
+        if (config.pantalaimon.use && !config.experimentalRustCrypto) {
             const pantalaimon = new PantalaimonClient(config.homeserverUrl, storage);
             client = await pantalaimon.createClientWithCredentials(config.pantalaimon.username, config.pantalaimon.password);
+        } else if (config.experimentalRustCrypto) {
+            if (config.pantalaimon.use) {
+                console.warn("You have a pantalaimon config activated and experimentalRustCrypto. Make sure the accessToken is set and pantalaimon is disabled!");
+            }
+            // 0 means sqlite. It comes from "@matrix-org/matrix-sdk-crypto-nodejs" and is ensured to be 0 by https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/blob/8cfee331a7fbfb00625bd3b86a78686a0f954534/tests/machine.test.js#L27-L31
+            const cryptoStorage = new RustSdkCryptoStorageProvider(path.join(storagePath, "crypto"), 0);
```
Gnuxie, line 81:
```suggestion
            const cryptoStorage = new RustSdkCryptoStorageProvider(path.join(storagePath, "crypto"), StoreType.Sqlite);
```

src/index.ts
```diff
@@ -69,9 +70,17 @@ import { initializeSentry, patchMatrixClient } from "./utils";
         const storage = new SimpleFsStorageProvider(path.join(storagePath, "bot.json"));
 
         let client: MatrixClient;
-        if (config.pantalaimon.use) {
+        if (config.pantalaimon.use && !config.experimentalRustCrypto) {
             const pantalaimon = new PantalaimonClient(config.homeserverUrl, storage);
             client = await pantalaimon.createClientWithCredentials(config.pantalaimon.username, config.pantalaimon.password);
+        } else if (config.experimentalRustCrypto) {
+            if (config.pantalaimon.use) {
+                console.warn("You have a pantalaimon config activated and experimentalRustCrypto. Make sure the accessToken is set and pantalaimon is disabled!");
```
Gnuxie, line 78: Why should we only warn? What would be the use case for saying use pantalaimon but we're gonna silently not use it and use matrix-sdk-crypto-nodejs?

src/config.ts
```diff
@@ -129,6 +129,9 @@ export interface IConfig {
             enabled: boolean;
         }
     }
+    // Experimental usage of the matrix-bot-sdk rust crypto.
+    // This can not be used with Pantalaimon.
+    experimentalRustCrypto: boolean;
```
Gnuxie, line 134: It would be good to document this in `config/default.yaml` too

src/index.ts
MTRNord: My reasoning was that this might have just been a leftover setting which doesnt change the outcome. So its more a "hey maybe cleanup your config" warning.

Gnuxie: I think we should refuse to start if they do that because they could not see the warning

src/index.ts
```diff
@@ -69,9 +70,17 @@ import { initializeSentry, patchMatrixClient } from "./utils";
...
         let client: MatrixClient;
-        if (config.pantalaimon.use) {
+        if (config.pantalaimon.use && !config.experimentalRustCrypto) {
             const pantalaimon = new PantalaimonClient(config.homeserverUrl, storage);
             client = await pantalaimon.createClientWithCredentials(config.pantalaimon.username, config.pantalaimon.password);
+        } else if (config.experimentalRustCrypto) {
+            if (config.pantalaimon.use) {
+                throw Error("You have a pantalaimon config activated and experimentalRustCrypto. Make sure the accessToken is set and pantalaimon is disabled!");
+            }
+            // 0 means sqlite. It comes from "@matrix-org/matrix-sdk-crypto-nodejs" and is ensured to be 0 by https://github.com/matrix-org/matrix-rust-sdk-crypto-nodejs/blob/8cfee331a7fbfb00625bd3b86a78686a0f954534/tests/machine.test.js#L27-L31
```
Gnuxie, line 80: Comment is now out of date surely?

config/default.yaml
```diff
@@ -28,6 +28,13 @@ pantalaimon:
   # After successfully logging in once, this will be ignored, so this value can be blanked after first startup.
   password: your_password
 
+# Experimental usage of the matrix-bot-sdk rust crypto.
+# This can not be used with Pantalaimon.
+# Make sure to setup the bot as if you are not using pantalaimon for this.
+#
+# Warning: At this time this is not considered production safe.
```
Gnuxie, line 35: Really? Are we saying that or are the authors of `RustSdkCryptoStorageProvider` saying that?

MTRNord: Copied from chat:

> I am saying that one. Its not well tested at all currently so I think it makes sense to first have this as a non prod flag. Also I remember occasional breaking stuff from hookshot users. So I think it makes sense to first have a release or 2 where this is available but not yet guaranteed to be stable since we didnt test it enough imho. And then we can revisit this and drop that sentence I guess.
> 
> Like I am seeing draupnir as something that should run reliably most of the time if not all of the time. And I dont think we have enough data if this is the case with bot-sdk (Though if
> [TravisR](https://matrix.to/#/@travis:t2l.io) thinks this is stable enough then I am also happy to remove the sentence. Its just that I cant with confidence attest this yet since I didnt test it extensively enough :) )

Gnuxie: https://matrix.to/#/!IaWNErZAgQUhGqJXjX:matrix.org/$hV8HlsBuarqdS7guacNp4Sek8OGGrpf4uqCMSJwx_Qs?via=matrix.org&via=envs.net&via=the-apothecary.club

MTRNord: Oh and link for future readers: https://matrix.to/#/!IaWNErZAgQUhGqJXjX:matrix.org/$hV8HlsBuarqdS7guacNp4Sek8OGGrpf4uqCMSJwx_Qs?via=matrix.org&via=envs.net&via=the-apothecary.club

src/index.ts
MTRNord: Oh right

Issue #176, opened 2023-11-27
Consider changing license to AFL-3.0


Gnuxie:
https://opensource.org/license/afl-3-0-php/

#### Why?

We are seeking a license similar to Apache 2.0 that allows for straight forward simple incorporation into other work but we don't want references to the original work to be buried away in a NOTICE file that potentially no one could ever see and be hard to track down the source of the original work. The AFL-3.0 seems like it could offer this as explained below. 

<sub><sup>For further optional reading on the license situation, if you want to discuss why we derived these requirements then please read the following. Yes it starts out dramatically and will make you recoil in the first few lines. Just sit tight and read to the end if you can manage it, if you can't, and want to debate these requirements, then you will not be welcome to and we're going to ask you right now to please go away. https://github.com/the-draupnir-project/Draupnir/blob/main/CONTRIBUTING.md#further-notes-on-license-and-its-relation-to-business-in-general<sub><sup>

##### Condition 5 means network interaction involves distribution of the work (by whoever has deployed it) (and then use by anyone who uses it). 

>5) External Deployment. The term “External Deployment” means the use, distribution, or communication of the Original Work or Derivative Works in any way such that the Original Work or Derivative Works may be used by anyone other than You, whether those works are distributed or communicated to those persons or made available as an application intended for use over a network. As an express condition for the grants of license hereunder, You must treat any External Deployment by You of the Original Work or a Derivative Work as a distribution under section 1(c).

##### Condition 3 means that the original source code (or a URL or whatever) must be provided whenever "Licensor" distributes the work.

>3) Grant of Source Code License. The term “Source Code” means the preferred form of the Original Work for making modifications to it and all available documentation describing how to modify the Original Work. Licensor agrees to provide a machine-readable copy of the Source Code of the Original Work along with each copy of the Original Work that Licensor distributes. Licensor reserves the right to satisfy this obligation by placing a machine-readable copy of the Source Code in an information repository reasonably calculated to permit inexpensive and convenient access by You for as long as Licensor continues to distribute the Original Work.

Ideally, anyone who distributes, creates a derivative work, or incorporates the original work into another would have to uphold condition 3 (with reference to the original work, not the source of the derivative). But it's made unclear whether this is an obligation as it seems to only apply to "Licensor", and I don't know whether someone who distributes the work or creates a derivative or incorporates the original work becomes a licensor too or not? Help me somebody

[...]

PR #204, opened 2024-01-14
Change license to AFL-3.0 [Contributors please read & respond]

Gnuxie:
Please reply with 👍 or approve this PR if you agree to relicense your work under the [Academic Free License](https://opensource.org/license/afl-3-0-php/)

As discussed in #176 we want to change the license to AFL-3.0 so that we have an approved open source license with a condition for distribution or incorporation to provide the source code for the original work (or a link to it and this is implemented currently via the status command. I think we should go a step further and provide the notices file here too as a follow up). 

Contributors previously were asked to agree to provide their [contribution](https://github.com/the-draupnir-project/Draupnir/blob/56e03e9/CONTRIBUTING.md#contributing-code-to-draupnir) under the Apache-2.0 license so that I would be the sole author contributing under the project's [Cooperative Software License](https://github.com/the-draupnir-project/Draupnir/blob/56e03e9/LICENSE), which would in (dubious) theory make relicensing to anything Apache-2.0 compatible straight forward later on. However out of an abundance of caution I would like all the authors who have contributed since forking (to cover the case that someone believes they have contributed under the CSL and not Apache-2.0, and therefore can object) to agree to the license change. This also leaves a concrete paper trail in one place rather than each PR.

Contributors:

- [x] @Mikaela 
- [x] @MTRNord 
- [x] @julianfoad 
- [x] @FSG-Cat 
- [x] @JokerGermany 
- [x] @progval

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,19 +1,5 @@
 # Contributing code to Draupnir
 
-Draupnir is licensed under the Cooperative Software License.
-The reason for the license is simply because this project
-was something that I was previously employed to work on.
-
-I have not decided whether I will accept contributions under the
-same license or another license or a CLA yet.
-I will welcome advice, but I will not welcome attempts to parsuade
-me to re-adopt and relicense under Apache-2.0 unless being offered
-compensation.
-
-As of now, I am accepting contributions under the Apache-2.0 license
-in the same way as Mjolnir. This allows me the option to relicense
-Draupnir under Apache-2.0 without needing to chase up all contributors.
-
 ## How to contribute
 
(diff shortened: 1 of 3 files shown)


FSG-Cat (approved):
I Catalan Lover approves of relicencing all my contributions to Draupnir to AFL-3.0 as this PR is suggesting at the time of writing. To the best of my knowledge all my contributions have been made under Apache 2.0 but if any have been done under other licences they are covered.

progval (approved):

JokerGermany (approved):

Mikaela (approved):

julianfoad (approved):

PR #221, opened 2024-01-25
MPS backend for Draupnir

Gnuxie:
#### This is a rewrite of the core functionality of Draupnir

This has been done with a library called the [matrix-protection-suite](https://github.com/Gnuxie/matrix-protection-suite) that Draupnir now depends upon. The protection suite provides a plugin system that allows all of Draupnir's functionality to be implemented as protections, that can be disabled/enabled and configured individually as distinct modules. The protection suite also introduces capability interfaces and capability providers which allow the user to modify what each individual protection can do without needing to change modify the protection's code.

#### Things that need to happen before merge:

- [x] Account data migration is verified in tests (either in MPS or here). https://github.com/Gnuxie/matrix-protection-suite/issues/7
- [x] Tests for member ban sychronisation and server ACL synchronisation from Draupnir are moved to MPS to test the new protections that take these responsibilities.
- [ ] Tests involving batch additions to policy rooms are moved to MPS, since the logic for those has also been moved there.
- [ ] Appropriate batching behaviour is added to MPS if they fail
- [x] Someone verifies that redaction integration tests exist within Draupnir, and that the appropriate behaviour is still provided by either Draupnir or MPS.  
- [ ] <del>The test for applying consequences in rooms by the most recently active room first still exists</del> this seems to only matter for server ACL application
- [x] Any other tests that are broken in Draupnir are fixed or moved to MPS.
- [x] Make a decision on whether persisting the room state cache and updating it more liberally needs to be an option before we make the MPS backend available to testing on the `develop`. The use case for this being people who run Draupnir against a homeserver that is very slow to respond e.g. matrix.org. 
- [ ] Ensure that `config.disableServerACL` disables the new `ServerPolicySynchronisationProtection` and otherwise it is enabled by default. Deprecate this config value at the same time.
- [x] Ensure tests for the MembershipRevision exists, particularly around redactions, we wouldn't want members to be silently dropped out of a room because redaction handling code is naive and unaware of redaction rules.
- [ ] If they don't exist: Tests for policy list revision. In particular because there are ways of indexing rules specific to the policy list revision called "scopes", and we need to test that they intern/unintern rules correctly based on state changes.
- [x] Mjolnir's old ReportManager and WebAPIs (for receiving reports) are enabled in some glue code that sets up bot mode only. Follow up issues are made to make these modules protections. 
- [ ]  Comprehensive list of changes is collected :( 
- [x]  SHORTCODES


Gnuxie:
CI Errors seem to be [this](https://github.com/yarnpkg/yarn/issues/6312#issuecomment-424334870) there's some comments about a case of using git dependencies that have a prepare script, like us. Lots of terrible suggestions in the thread. Don't understand. https://github.com/yarnpkg/yarn/issues/6312#issuecomment-469706434 is probably true, the ubuntu/rust dev image probably uses an ancient version of yarn but need to confirm. But that will only elaborate on the error.

Gnuxie:
I also don't know why pre-commit hasn't been running correctly locally

Gnuxie:
>v1.22.21

> Upgrade to 1.13+ to get the true error 

Thanks buddy i got the true one for real

Gnuxie:
It's probably my fault for using `git+http` and `prepare` wrong.

Gnuxie:
I highly suspect that `getFirstEventMatching` can race and we can get a timeline event down `/sync` before we finish awaiting for a response to the http call that sent it.

Gnuxie:
Duplicate `RoomStateRevisionIssuers` are somehow being created, and only one is being updated while the other is being given to membership revision issuers

Gnuxie:
Yeah we never fixed a race condition in the `InternedInstanceFactory` where the instance factory can get called concurrently...

Issue #210, opened 2024-01-19
Recieving a new policy from a watched list (e.g. a consumer of CME) doesn't cause a redaction until someone joins/leaves the room again


Gnuxie:
A spammer joined fluffychat's support room.
I issued a CME ban via TNO
fluffy consumed the ban and banned the user, but didn't redact the messages because nothing triggered the redaction queue
i had to force the queue to trigger by leaving and rejoining in that room. https://github.com/the-draupnir-project/Draupnir/blob/main/src/ProtectedRoomsSet.ts#L216

Mikaela:
Easier workaround /myroomnick was just tested successfully in the Draupnir room.

PR #230, opened 2024-02-01
Stop processing redaction queue before bans.

Gnuxie:
Introduced in
https://github.com/the-draupnir-project/Draupnir/commit/ea49b0a5363861b3729dcc720ff091b2480c4165. Quite easy to see how.
Future redaction work should unit test for this.
This can be done quite easily by using a setTimeout, and non hackily if you can make the fake request to ban a user resolve only when an item in the redaction queue has been processed.

Fixes https://github.com/the-draupnir-project/Draupnir/issues/210.


Gnuxie:
Follow up https://github.com/the-draupnir-project/Draupnir/issues/231

PR #242, opened 2024-02-07
Read secrets from command line provided paths, if specified

TheArcaneBrony:


src/config.ts
```diff
@@ -313,3 +305,54 @@ export const SOFTWARE_VERSION = (() => {
...
+ * @param arg Argument name
+ * @param throwOnInvalid If true or undefined, throws an error if the argument is present but has no value. If false, returns undefined.
+ * @returns value passed to the argument or undefined.
+ * @throws Error if the argument is present but has no value.
+ */
+function getCommandlineStringArgument(arg: string, throwOnInvalid?: boolean | undefined): string | undefined {
+    // we don't want to throw if the argument is not present
+    if(!getCommandlineArgumentPresent(arg)) return undefined;
+
+    const index = process.argv.indexOf(arg);
```
Gnuxie, line 332: The reason why previously this was previously called `configOptionIndex` is so that it is explicit to the reader what the index is for. Please don't erase verbose names like this, they are there for a reason.

src/config.ts
```diff
@@ -313,3 +305,54 @@ export const SOFTWARE_VERSION = (() => {
...
+}
+
+/**
+ * Grabs a path argument from the command line and checks if it exists.
+ * @param arg Argument name
+ * @param throwOnInvalid If true or undefined, throws an error if the path does not exist. If false, returns undefined.
+ * @returns Path if it exists, otherwise undefined.
+ * @throws Error if the path does not exist and throwOnInvalid is true or undefined.
+ */
+function getCommandlinePathArgument(arg: string, throwOnInvalid?: boolean | undefined): string | undefined {
```
Gnuxie, line 349: Can we replace all instances of `Commandline` with `CommandLine` please?

Gnuxie (commented):
Thanks for the PR, I have some small style issues for now, working on reviewing the rest

src/config.ts
```diff
@@ -313,3 +305,54 @@ export const SOFTWARE_VERSION = (() => {
...
+ */
+function getCommandlineStringArgument(arg: string, throwOnInvalid?: boolean | undefined): string | undefined {
+    // we don't want to throw if the argument is not present
+    if(!getCommandlineArgumentPresent(arg)) return undefined;
+
+    const index = process.argv.indexOf(arg);
+    if (index === -1) return undefined;
+    //check if the next index is not an argument
+    if (process.argv[index + 1] && !process.argv[index + 1].startsWith("--"))
+        return process.argv[index + 1];
```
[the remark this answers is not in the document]
TheArcaneBrony: Ah, I personally don't use braces for single line statements since it increases reading difficulty for me, but I'll fix this up :)

src/config.ts
```diff
@@ -224,34 +224,26 @@ export function getDefaultConfig(): IConfig {
...
-            throw new Error("No path provided with option --draupnir-config");
-        }
-        return configOptionPath;
-    } else {
-        return;
-    }
-}
-
 export function read(): IConfig {
-    const explicitConfigPath = configPathFromArguments(process.argv);
```
Gnuxie, line 246: You shouldn't have erased explicitly passing the arguments vector to this function. It was done so that it would be possible to test the function.

TheArcaneBrony: Can you really attach tests to non-exported functions though?

Gnuxie: Good observation, but it's still better to use explicit arguments over global variables where possible. It will make things easier for the future.

src/config.ts
```diff
@@ -313,3 +305,54 @@ export const SOFTWARE_VERSION = (() => {
...
+// Command line related functions
+
+/**
+ * Grabs an argument from the command line and checks if it exists.
+ * @param arg Argument name
+ * @returns True if the argument is present, otherwise false.
+ */
+function getCommandlineArgumentPresent(arg: string): boolean {
+    return process.argv.includes(arg);
+}
```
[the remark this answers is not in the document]
Gnuxie: It could be, but I'd like to at the least follow an authoritative source and use their terminology when referring to command line arguments. If you can find something more appropriate then please let me know. I am not familiar with interfacing with the OS or shell or wherever the arguments array actually comes from. 

If you think we should be even more specific by calling them `longOptionNames` then I'd agree.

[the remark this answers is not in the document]
TheArcaneBrony: Honestly, I was thinking of renaming them to eg. `getNonPOSIXCommandLineStringLongOption`, just to be clear that it's not fully compliant (no `=`)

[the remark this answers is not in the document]
TheArcaneBrony: (Tiny infodump: "argv" comes from the POSIX C argument name for the argument list, passed along to `execv`/`execve`)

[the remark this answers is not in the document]
Gnuxie: `String` in the name is redundant, we're not parsing anything. There's no need to say it's not posix compliant, but I am genuinely interested to know where this convention came from, if we don't know it shouldn't block this PR.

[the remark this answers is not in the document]
TheArcaneBrony: I have no clue where the convention came from, nor am I able to find anything about it, but I know it's fairly common in shell scripts.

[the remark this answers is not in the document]
TheArcaneBrony: The intent for explicitly saying `String` was to differentiate it from `Path`, by the way.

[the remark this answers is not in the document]
TheArcaneBrony: As for the option vs long option thing, I don't think it really matters since the way it's implemented perfectly allows you to do eg. `-c /some/path` if you wanted short options, since the dashes are passed in the argument name.

src/config.ts
```diff
@@ -313,3 +307,70 @@ export const SOFTWARE_VERSION = (() => {
...
+ */
+function getCommandLineStringOption(args: string[], arg: string, throwOnInvalid?: boolean): string | undefined {
+    // We don't want to throw if the option is not present
+    if(!isCommandLineOptionPresent(args, arg)) {
+        return undefined;
+    }
+
+    const optionIndex = args.indexOf(arg);
+    if (optionIndex === -1) {
+        return undefined;
```
Gnuxie, line 338: Both of these checks are surely redundant, we only need the second one right?

src/config.ts
```diff
@@ -313,3 +307,70 @@ export const SOFTWARE_VERSION = (() => {
...
+    }
+
+    //check if the next index is not an option
+    if (args[optionIndex + 1] && !args[optionIndex + 1].startsWith("--")){
+        return args[optionIndex + 1];
+    }
```
Gnuxie, line 344: Good

src/config.ts
```diff
@@ -313,3 +307,70 @@ export const SOFTWARE_VERSION = (() => {
...
+}
+
+/**
+ * Grabs an option's value from program options if it exists, otherwise returns undefined.
+ * @param arg Option name
+ * @param throwOnInvalid If true or undefined, throws an error if the option is present but has no value. If false, returns undefined.
+ * @returns The value passed to the option, or undefined if the option is not specified.
+ * @throws Error if the option is present but has no value.
+ */
+function getCommandLineStringOption(args: string[], arg: string, throwOnInvalid?: boolean): string | undefined {
```
Gnuxie, line 330: We should remove `throwOnInvalid` entirely for this function, if they want to check for a flag, they can just use `isCommandLineOptionPresent` instead.

src/config.ts
```diff
@@ -313,3 +307,70 @@ export const SOFTWARE_VERSION = (() => {
...
+
+
+// Command line related functions
+
+/**
+ * Grabs an option from the command line and checks if it exists.
+ * @param arg Option name
+ * @returns True if the option is present, otherwise false.
+ */
+function isCommandLineOptionPresent(args: string[], arg: string): boolean {
```
Gnuxie, line 319: can we rename all instances of `arg` to `optionName` please?

src/config.ts
```diff
@@ -313,3 +307,70 @@ export const SOFTWARE_VERSION = (() => {
...
+        throw new Error(`Invalid value provided for ${arg}`);
+    }
+
+    return undefined;
+}
+
+/**
+ * Grabs a path option from the command line and checks if it exists.
+ * @param arg Option name
+ * @param throwOnInvalid If true or undefined, throws an error if the path does not exist. If false, returns undefined.
```
Gnuxie, line 356: `throwOnInvalid` makes less sense here. If someone wanted to set this explicitly to false, so they could use the path as a location to write to (given we have included a check to see if the path exists here), then this function will not throw if a path wasn't provided at all.

Gnuxie: If it was included so that someone could provide a path to write to instead of read from, then i suggest we just remove the check for `fs.existsSync` entirely. And leave it to the caller to verify if the path exists.

src/config.ts
```diff
@@ -224,34 +224,28 @@ export function getDefaultConfig(): IConfig {
...
-        if (!configOptionPath) {
-            throw new Error("No path provided with option --draupnir-config");
-        }
-        return configOptionPath;
-    } else {
-        return;
-    }
-}
-
 export function read(): IConfig {
```
Gnuxie, line 227:
```
/**
 * @returns The users's raw config, deep copied over the `defaultConfig`.
 */
function readConfigSource(): IConfig {
    const explicitConfigPath = getCommandLinePathOption(process.argv, "--draupnir-config", true);
    if (explicitConfigPath !== undefined) {
        const content = fs.readFileSync(explicitConfigPath, "utf8");
        const parsed = load(content);
        return Config.util.extendDeep({}, defaultConfig, parsed);
    } else {
        return Config.util.extendDeep({}, defaultConfig, Config.util.toObject()) as IConfig;
    }
}

export function read(): IConfig {
    const config = readConfigSource();
    const explicitAccessTokenPath = getCommandLinePathOption(process.argv, "--access-token-path", true);
    const explicitPantalaimonPasswordPath = getCommandLinePathOption(process.argv, "--pantalaimon-password-path", true);
    if (explicitAccessTokenPath !== undefined) {
        config.accessToken = fs.readFileSync(explicitAccessTokenPath, "utf8");
    }
    if (explicitPantalaimonPasswordPath) {
        config.pantalaimon.password = fs.readFileSync(explicitPantalaimonPasswordPath, "utf8");
    }
    return config;
}
```

How about something like this?

src/config.ts
TheArcaneBrony: I would remove the second one, personally (consistency with the path function, and i'd say its fair to assume optionIndex would never be -1 if it's present, right?)

src/config.ts
TheArcaneBrony: It's there to catch passing no parameter, and consistency

src/config.ts
TheArcaneBrony: Should I invert the logic and rename it to `checkPathExists`?

src/config.ts
TheArcaneBrony: Guessing the intent is to separate concerns between functions?

Gnuxie: Really it was to remove the `var` for config (which should have also been a `let`). Another style choice in Draupnir that should be documented is avoiding `let` wherever possible (but there are still reasons to use it sometimes).  But yeah i guess.

src/config.ts
Gnuxie: I think the logic should be removed entirely, surely we will choke if the path doesn't exist when someone tries to read it, which in our case isn't very far down the line.

src/config.ts
Gnuxie: Well, it does make me feel uneasy, since we are using `optionIndex`. If the implementation for `isCommandLineOptionPresent` breaks, then our now naive use of `optionIndex` will be confusing to whoever has to discover it is now `-1`.

src/config.ts
Gnuxie: It's not a good thing to be consistently confusing.

TheArcaneBrony:
Code has been tested for a few cases (passing no arg, passing invalid path, not passing arg at all, and passing a valid file), appears to work as expected

src/config.ts
```diff
@@ -313,3 +311,38 @@ export const SOFTWARE_VERSION = (() => {
     // into <pre> or <code> where it will create an unnecessary newline.
     return /^(.*)$/m.exec(versionFile)?.at(0) ?? defaultText;
 })();
+
+
+// Command line related functions
+
+/**
+ * Grabs an option from the command line and checks if it exists.
+ * @param arg Option name
```
Gnuxie, line 320: this is called `arg` and not `optionName`

src/config.ts
```diff
@@ -225,35 +225,33 @@ export function getDefaultConfig(): IConfig {
 }
 
 /**
- * Grabs an explicit path provided for mjolnir's config from an arguments vector if provided, otherwise returns undefined.
- * @param argv An arguments vector sourced from `process.argv`.
- * @returns The path if one was provided or undefined.
+ * @returns The users's raw config, deep copied over the `defaultConfig`.
```
Gnuxie, line 228: Was this supposed to be deleted? not sure if the diff is just being weird here though

Gnuxie (approved):

src/config.ts
TheArcaneBrony: Diff being weird I think, those "removed" lines are from the old "get config path" function

Gnuxie: oki

PR #272, opened 2024-02-29
Update D4A Documentation to be A LOT better. 

FSG-Cat:
The old D4A documentation was omiting a fair number of things one needed to know and also tbh didnt even explain half of the stuff you would want to know and it makes sense. Alpha docs from a very rushed time. Well today lets fix that as i upgraded the docs as part of my work of upstreaming what was learned in the creation of https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/3204


docs/appservice.md
```diff
@@ -2,31 +2,53 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
 This module is currently alpha quality and is subject to rapid changes,
 it is not recommended currently and support will be limited.
 
+Usage of the [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md) role for Draupnir for all is recommended above trying to deploy Draupnir for all your self from scratch as much more support is offered when taking this route. Especially if not deviating from defaults.
```
Gnuxie, line 5: nit: replace "above" with "instead of" so not to be confused.

Pick either "your self" or "from scratch".

src/appservice/config/config.production.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - "DoLoR"
+      - "aMeT"
+
+    # For how long (in minutes) the user is "new" to the WordList plugin.
+    #
+    # After this time, the user will no longer be banned for using a word in the above wordlist.
+    #
+    # Set to zero to disable the timeout and make users *always* appear "new".
+    # (users will always be banned if they say a bad word)
+    minutesBeforeTrusting: 20
```
Gnuxie, line 108: Hmm why did you duplicate this and not just tell them to reuse the one in `config/`?

FSG-Cat: Its a subset of the one in config/ that only includes keys that are actually inherited. Well technically thats a lie verbose is not mentioned at all

Gnuxie: Hmm, i wish there was a way to do this without duplicating since the `default.yaml` kind of is the documentation, which tbh also sucks.

FSG-Cat: Tbh i think that there is a better way to do this in the future but i would say lets leave that for later. Since i think this is a case where perfect is the enemy of good. Since if we want to be perfect the whole AS stuff should also be in the config dir.

Gnuxie: Perfect is the enemy of good is gonna go on the list of banned expressions. Since, that's the kind of dogma that has driven the ecosystem down the drain. 

Regardless, I'd rather delete the file and tell them which keys will be copied across.

docs/appservice.md
Gnuxie: why'd you mark this as resolved and undo the change you made?

docs/appservice.md
```diff
@@ -2,7 +2,7 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
 This module is currently alpha quality and is subject to rapid changes,
 it is not recommended currently and support will be limited.
 
-Usage of the [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md) role for Draupnir for all is recommended instead of trying to deploy Draupnir for all from scratch as much more support is offered when taking this route. Especially if not deviating from defaults.
+Usage of the [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md) role for Draupnir for all is recommended above trying to deploy Draupnir for all your self from scratch as much more support is offered when taking this route. Especially if not deviating from defaults.
```
Gnuxie: i dno't think you were meant to do this

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
 This module is currently alpha quality and is subject to rapid changes,
 it is not recommended currently and support will be limited.
 
+Usage of the [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md) role for Draupnir for all is recommended instead of trying to deploy Draupnir for all from scratch as much more support is offered when taking this route. Especially if not deviating from defaults.
```
Gnuxie, line 5: Not important but we should really have a column limit in `.editorconfig` or something

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
...
+
+8. Copy the `draupnir-registration.yaml` to your matrix homeserver and refer to it in `homeserver.yaml` like so:
 ```
   app_service_config_files:
     - "/data/draupnir-registration.yaml"
 ```
+
+# Post Install Setup and Configuration.
+
+If you successfully followed Steps 1-8 or got to the point your running the bot thru another mean congratulations. Time for Post Install Finalisation and Configuration of your new appservice.
```
Gnuxie, line 53: `thru`

Gnuxie: >Time for Post Install Finalisation and Configuration of your new appservice.

Then there are some post install steps and configuration that you should apply to your new appservice

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
...
+
+1. If your successfully at this point it’s assumed that your `draupnir-moderation` or `draupnir-main` bot is in the room with `$MANAGEMENT_ALIAS` as it’s alias so make sure it’s in there before proceeding.
+
+2. Make your Appservice Management room private as it should be.
```
Gnuxie, line 57: Set your appservice management room to private, if you didn't already. Anyone with access to this room can manage access to the appservice, so be careful who you share this with.

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
...
 ```
   app_service_config_files:
     - "/data/draupnir-registration.yaml"
 ```
+
+# Post Install Setup and Configuration.
+
+If you successfully followed Steps 1-8 or got to the point your running the bot thru another mean congratulations. Time for Post Install Finalisation and Configuration of your new appservice.
+
+1. If your successfully at this point it’s assumed that your `draupnir-moderation` or `draupnir-main` bot is in the room with `$MANAGEMENT_ALIAS` as it’s alias so make sure it’s in there before proceeding.
```
Gnuxie, line 55: is in the admin room (the room with designated by `$MANAGEMENT_ALIAS`). So go back up and follow those setps if you haven't got there yet.

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
...
+
+2. Make your Appservice Management room private as it should be.
+
+3. Decide if you want to allow provisioning per homeserver or per user. If you choose to only provision per user skip to step 5.
+
+4. Allow provisioning per homeserver. To provision per homeserver you write in your control room /plain MXID_OF_APPSERVICE_HERE allow @*:homeserver_to_allow
+
+5. Allow provisioning per user. To Provision per user you write in your control room /plain MXID_OF_APPSERVICE_HERE allow MXID_TO_ALLOW
+FIXME: If the client is being dumb and adding escapes to lone instances of * in the command strip them out so you don't have to mandate /plain use to guarantee the client doesn’t screw the user over.
```
Gnuxie, line 64: hmmm

brknkfr:
Maybe add somewhere in the beginning, that the appservice is only working when homeserver allows the creation of an unencrypted management room.

docs/appservice.md
```diff
@@ -2,31 +2,65 @@ Draupnir can be run as an appservice, allowing users you trust or on your homese
 This module is currently alpha quality and is subject to rapid changes,
 it is not recommended currently and support will be limited.
 
+Usage of the [matrix-docker-ansible-deploy](https://github.com/spantaleev/matrix-docker-ansible-deploy/blob/master/docs/configuring-playbook-bot-draupnir.md) role for Draupnir for all is recommended instead of trying to deploy Draupnir for all from scratch as much more support is offered when taking this route. Especially if not deviating from defaults.
+
 # Prerequisites
 
 This guide assumes you will be using Docker and that you are able to provide a postgres database for Draupnir to connect to in application service mode.
```
FSG-Cat, line 10:
```suggestion

Please note that Draupnir in appservice mode does not support E2EE nor support use of an E2EE proxy like [pantalaimon](https://github.com/matrix-org/pantalaimon) as there is currently no proxy for appservices like there is for /sync

```

PR #284, opened 2024-03-07
Fix d4all protections

MTRNord:
This fixes a long standing bug where we had things activated that didnt seem active. This should now have a clean protection instance per mjolnir instance.

### DIFF
--- a/src/protections/ProtectionManager.ts
+++ b/src/protections/ProtectionManager.ts
@@ -45,18 +45,6 @@ import { MatrixDataManager, RawSchemedData, SchemaMigration, SCHEMA_VERSION_KEY 
 import { Permalinks } from "../commands/interface-manager/Permalinks";
 import { CommandExceptionKind } from "../commands/interface-manager/CommandException";
 
-const PROTECTIONS: Protection[] = [
-    new FirstMessageIsImage(),
-    new BanPropagation(),
-    new BasicFlooding(),
-    new WordList(),
-    new MessageIsVoice(),
-    new MessageIsMedia(),
-    new TrustedReporters(),
-    new DetectFederationLag(),
-    new JoinWaveShortCircuit(),
-];
-
 const ENABLED_PROTECTIONS_EVENT_TYPE = "org.matrix.mjolnir.enabled_protections";
 type EnabledProtectionsEvent = RawSchemedData & {
     enabled: string[],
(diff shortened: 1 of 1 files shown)


src/protections/ProtectionManager.ts
```diff
@@ -146,9 +135,21 @@ export class ProtectionManager {
     get protections(): Readonly<Map<string /* protection name */, Protection>> {
         return this._protections;
     }
+    private readonly protection_classes: Protection[] = [
```
Gnuxie, line 138: NIT: Let's stick to the naming convention of `protectionClasses` please, sorry if that's been butchered in some other place.

src/protections/ProtectionManager.ts
```diff
@@ -81,9 +65,14 @@ class EnabledProtectionsManager extends MatrixDataManager<EnabledProtectionsEven
     private readonly enabledProtections = new Set</* protection name */string>();
 
     constructor(
-        private readonly mjolnir: Mjolnir
+        private readonly mjolnir: Mjolnir,
+        private readonly protections: Protection[]
     ) {
         super()
+        const banPropagationProtection = this.protections.find(p => p.name === 'BanPropagationProtection');
```
Gnuxie, line 72: would be cool if we could make 'BanPropagationProtection' a static `name` field on the BanPropagationProtection class so we don't have to risk typoing it. Which would then remove the need for this check

src/protections/ProtectionManager.ts
Gnuxie, line 138: Hmm seems like the `_protections` member is already doing this job...

src/protections/ProtectionManager.ts
```diff
@@ -159,7 +160,7 @@ export class ProtectionManager {
         await this.enabledProtectionsManager.start();
         this.mjolnir.reportManager.on("report.new", this.handleReport.bind(this));
         this.mjolnir.matrixEmitter.on("room.event", this.handleEvent.bind(this));
-        for (const protection of PROTECTIONS) {
+        for (const protection of this.protection_classes) {
```
Gnuxie, line 163: Maybe we could just move the list of protections (formerly) `PROTECTIONS` to here? and then pass `_protections` to the EnabledProtectionsManager`?

PR #301, opened 2024-03-20
[WIP] Find Command

ShadowJonathan:
This is a Find command from the Before Times, which im quickly just making a PR here for to at least give testament to it.

Basically, it would find users by regex given, and return which users it sees and knows about.

TODO:
- [ ] Convert this to TSX format, this could do with a threadifier a-la rules command
- [ ] Maybe make an addition to this, where it would also refer which rooms the user(s) are in? `--rooms`
- [ ] Have a way to specify looking for protected and/or all rooms that draupnir can see, currently it just does protected rooms

### DIFF
--- a/src/commands/CommandHandler.ts
+++ b/src/commands/CommandHandler.ts
@@ -41,6 +41,7 @@ import {
 import { execSetPowerLevelCommand } from "./SetPowerLevelCommand";
 import { execResolveCommand } from "./ResolveAlias";
 import { execKickCommand } from "./KickCommand";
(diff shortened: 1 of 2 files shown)


MTRNord:
How is this different from the `!draupnir rules matching <entity>` command? 🤔

ShadowJonathan:
Since (iirc) that only matches which rules are currently active, and searching through that.

This would look and find users matching a certain regex, which can be useful for figuring out spambot kick/ban usernames before hitting that command

PR #387, opened 2024-05-04
Documentation Website Move

FSG-Cat:
This PR changes over our documentation to the documentation website with the exception being the Synapse module for now.


Gnuxie (commented):
image

We should probably make the headers consistent first and figure out the order

FSG-Cat:
True that we should probably improve the site more. This PR can sit and wait until then. But lets atleast put a freeze on any more changes to docs in main.

Gnuxie (approved):

PR #388, opened 2024-05-04
Update matrix-appservice-bridge

MTRNord:
This updates to the latest versions of matrix-appservice-bridge and as a result also updates the bot-sdk.

The added `"matrix-bot-sdk": "npm:@vector-im/matrix-bot-sdk@^0.7.0-element.0",` is to replace the removed upstream alias. It does the exact same thing, but is now in draupnir instead.

**Warning: This also bumps the node version required to 20 due to the [support policy over at matrix-appservice-bridge](https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/docs/platforms.md#node-version-support), demonstrated [here](https://github.com/matrix-org/matrix-appservice-bridge/pull/487).**

As a result, we would drop debian support as only debian experimental seems to have version 20 at this time

### DIFF
--- a/.github/workflows/mjolnir.yml
+++ b/.github/workflows/mjolnir.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Specifically use node 18 like in the readme.
       uses: actions/setup-node@v4
       with:
-        node-version: '18'
+        node-version: '20'
     - run: yarn install
     - run: yarn build
     - run: yarn lint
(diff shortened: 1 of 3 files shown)


Gnuxie:
Thanks, I'm going to hold off while I get some of the more problematic bugs fixed in `v2.0.0-beta.*` just in case someone on debian updated. But either way i will be using this PR sooner or later.

PR #391, opened 2024-05-04
Move Synapse module to new website

FSG-Cat:
The synapse module was forgotten in the previous move.


Gnuxie:
wait why's this PR so huge?

Issue #416, opened 2024-05-13
Add `watch` prompt for when Draupnir is invited to a policy room.


Gnuxie:
related https://github.com/the-draupnir-project/Draupnir/issues/415

Issue #415, opened 2024-05-13
Add `rooms add` propagation protection prompt, to prompt to protect a room after being invited if it isn't already protected.


Gnuxie:
Do the same when being kicked from rooms by people in the management room.

Essentially this means that a Draupnir user doesn't even have to use the `rooms add` command, they can just invite the bot, set the permissions and click the prompt. We'd probably want to pair this with the "accept invites from space" config so that it is consistent with it and prompts whenever an authorized invitation is received.

PR #422, opened 2024-05-13
Invitation protections

Gnuxie:
- [x] Extract part of the `joinOnInviteListener`to common code that checks whether an invitation is authorized for the purpose of protecting or watching a list.
- [x] Replace the remaining portion of the join on invite listener with a protection that joins the rooms on invite.
- [x] Add a protection to prompt to protect rooms that Draupnir has been invited to by authorized users https://github.com/the-draupnir-project/Draupnir/issues/415
- [x] Add a protection to prompt to add the room as a watched list when it contains policies https://github.com/the-draupnir-project/Draupnir/issues/416 (or has the room type etc)
- [x] considerations: `config.protectAllJoinedRooms` should disable the protect rooms protection and explain why when it does that. 

Consider undoing all the above if we are kicked or leave.

Tied to https://github.com/the-draupnir-project/planning/issues/15


Gnuxie:
The current work is great, but we probably need to treat the `ProtectRoomsOnInviteProtection` as an entrypoint for something else. 

This entry point would just accept the invitation, we'd need to defer to something else to decide whether to prompt to protect the room or watch a list, since we don't have access to the room state until we join the room. We probably need to still ask to protect the room when it is a list.

Issue #453, opened 2024-06-15
Remove `root` from JSX in `DeadDocument`


Gnuxie:
So, it's fine that the root node exists in `DeadDocument`, but i don't think it should be placed explicitly into templates.

PR #454, opened 2024-06-15
TypeCheck JSX expressions, children, and attributes ^-^

Gnuxie:
Awesome.

The documentation was not very explicit, I had to play around
a lot to figure it out but I'm so happy that this works.
https://www.typescriptlang.org/docs/handbook/jsx.html.

Please note that `undefined` as a child will not be caught
yet, as we need to enable `exactOptionalPropertyTypes` for
this to work.

### DIFF
--- a/src/commands/interface-manager/JSXFactory.ts
+++ b/src/commands/interface-manager/JSXFactory.ts
@@ -6,7 +6,7 @@
 import { DocumentNode, LeafNode, makeDocumentNode, makeLeafNode, NodeTag, TextNode } from "./DeadDocument";
 import { DeadDocumentPresentationMirror } from "./DeadDocumentPresentation";
 
-type rawJSX = DocumentNode|LeafNode|string|number|Array<rawJSX>;
+type JSXChild = DocumentNode|LeafNode|string|number|JSXChild[];
 
 export function JSXFactory(tag: NodeTag, properties: unknown, ...rawChildren: (DocumentNode|LeafNode|string)[]) {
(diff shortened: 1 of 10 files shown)


src/commands/interface-manager/JSXFactory.ts
```diff
@@ -37,23 +37,51 @@ export function JSXFactory(tag: NodeTag, properties: unknown, ...rawChildren: (D
...
+            i: NodeProperties,
+            li: NodeProperties,
+            ol: NodeProperties & { 'start'?: number },
+            p: NodeProperties,
+            pre: NodeProperties,
+            root: NodeProperties,
+            span: NodeProperties & {
+                'data-mx-bg-color'?: string,
+                'data-mx-color'?: string,
+                'data-mx-spoiler'?: string | null
```
Gnuxie, line 76: hmm, this needs fixing to be `data-max-spoiler?: string | undefined` now that i discovered https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes. But I wonder if there is another way?

Gnuxie:
https://github.com/the-draupnir-project/Draupnir/issues/453
https://github.com/the-draupnir-project/Draupnir/issues/452

Gnuxie:
Hmmm we could possibly use https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts

Gnuxie:
We might be able to avoid `declare global` if we can try import the entire module in each tsx file and not just JSXFactory.

PR #471, opened 2024-07-01
Avoid invalid createRoom calls when making a list as the draupnir user

morguldir:
Signed-off-by: morguldir <morguldir@protonmail.com>


Gnuxie (approved):
Thank you so much <3

PR #477, opened 2024-07-17
Add SBOM and Attestation to the Docker release process

MTRNord:
Also ensure we only deploy platforms on the qemu action as needed and update versions. (However, it is expected that it might install more than specified as it relies on prebuilt bundles)

This also limits the permissions used a bit.

See https://github.com/actions/attest-build-provenance for more info on the attestations. SBOM is probably self-explanatory. 

Signed-off-by: MTRNord <bold.wing5824@fastmail.com>


.github/workflows/docker-hub-develop.yml
```diff
@@ -26,10 +31,12 @@ jobs:
 
       # Needed for multi platform builds
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3.0.0
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: arm64
```
Gnuxie, line 36: Just double checking but this doesn't mean that we only build for arm64 right?

.github/workflows/docker-hub-release.yml
```diff
@@ -38,11 +45,21 @@ jobs:
...
           tags: |
             ${{ env.DOCKER_NAMESPACE }}/draupnir:${{ env.RELEASE_VERSION }}
+
+      - name: Attest
+        uses: actions/attest-build-provenance@v1
+        id: attest
+        with:
+          subject-name: ${{ env.DOCKER_NAMESPACE }}/draupnir:${{ env.RELEASE_VERSION }}
+          subject-digest: ${{ steps.push.outputs.digest }}
+          push-to-registry: true
```
Gnuxie, line 65: how does the SBOM get made? does it take into account `yarn.lock`? Seems like https://github.com/anchore/syft gets [used by default](https://docs.docker.com/build/attestations/sbom/#sbom-generator) which is probably ok idk.

.github/workflows/docker-hub-develop.yml
MTRNord: No this just means it only needs to install the arm64 qemu stuff. It should natively build amd64 stuff still with this from how I understand the documentation.

.github/workflows/docker-hub-release.yml
MTRNord: Going to check that in a moment but it is supposed to do the sbom of the yarn stuff too. If not that would be worrying. It would be solvable by doing an sbom ourself and providing it to the docker image at `/usr/local/share/sbom/` but I would rather not need to do that.

MTRNord: Yes it reads the lockfile:

```
> syft . -q -o json | jq '.artifacts | map(select(.locations | first | select(.path == "/yarn.lock"))) | length'
579
```

Gnuxie: that's cool

.github/workflows/docker-hub-develop.yml
Gnuxie: guess we'll find out, but do you think we should state it explicitly just to be sure?

MTRNord: I can add a comment in a moment :)

MTRNord: Turns out you were right and this is wrong. Fixing it

.github/workflows/docker-hub-latest.yml
```diff
@@ -26,10 +31,12 @@ jobs:
 
       # Needed for multi platform builds
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3.0.0
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: ${{ env.PLATFORMS }}
```
Gnuxie, line 36: where are we defining this?

MTRNord: further up in the file(s) under the `env` key

Gnuxie:
Thank you so much <3

PR #486, opened 2024-07-23
Add reuse.software

Gnuxie:
We have a lot of verbose headers, and i think now is the best opportunity we have to become reuse compliant given that we just did two other similar maintenance changes (prettier, typescirpt5 & eslint9 & typescript-eslint).


.github/workflows/sign-off.yml
```diff
@@ -1,12 +1,10 @@
...
+# SPDX-FileCopyrightText: 2024 Gnuxie <Gnuxie@protonmail.com>
 #
 # SPDX-License-Identifier: Apache-2.0
-#
-# SPDX-FileAttributionText: <text>
-# This modified file incorporates work from matrix-org/backend-meta
-# https://github.com/matrix-org/backend-meta
-# </text>
+# SPDX-License-Identifier: CC0-1.0
+
```
Gnuxie, line 7: Whoopsie!!

Gnuxie, line 7: whoopsie!

Gnuxie:
Maybe we should try unnest the REUSE.toml's for the config directories and keep everything in the top level REUSE.toml. If it'll let us do that.

Gnuxie:
the top level file is quite noisy so i don't really wanna do that.

PR #536, opened 2024-09-06
Extract the `@the-draupnir-project/interface-manager` system to its own library where it can be better tested and improved. 

Gnuxie:


src/Draupnir.ts
```diff
@@ -291,11 +263,11 @@ export class Draupnir implements Client {
...
       );
-      await renderMatrixAndSend(
-        renderStatusInfo(statusInfo),
+      await sendMatrixEventsFromDeadDocument(
+        this.clientPlatform.toRoomMessageSender(),
         this.managementRoomID,
-        undefined,
-        this.client
+        renderStatusInfo(statusInfo),
+        {}
```
Gnuxie, line 270: status isn't being shown on startup anymore.

src/commands/interface-manager/MatrixHelpRenderer.tsx
```diff
@@ -326,3 +303,70 @@ export function renderMentionPill(
...
+  );
+  return (
+    <fragment>
+      <b>Documentation: </b> <a href={DOCUMENTATION_URL}>{DOCUMENTATION_URL}</a>
+      <br />
+      {[...groupedAndSortedCommands.entries()].map(
+        ([sourceTable, sortedEntries]) =>
+          renderSourceTableSummary(sourceTable, sortedEntries)
+      )}
+    </fragment>
```
Gnuxie, line 370: Subsequent tables are not being shown in a fragment

Gnuxie:
The current issue is that for some reason only the help command is being found, possibly because we're interning the draupnir table wrong.

PR #569, opened 2024-09-19
Start draupnir bots in batches when running in appservice mode

MTRNord:
This upstreams https://github.com/the-draupnir-project/Draupnir/blob/9a4f48c31e2289a2c4db308fd6f5d60c3a61fda4/src/appservice/MjolnirManager.ts#L237-L276 from my d4all branch. Idk if this is a too big or small chunk size but it seems to work in prod. Its a generator simply because it was giving a convenient iterator api to use. Though technically I am not making much use of it. So not sure if it should be like this but in the end its also not bad as a generator so idk.

### DIFF
--- a/src/appservice/AppServiceDraupnirManager.ts
+++ b/src/appservice/AppServiceDraupnirManager.ts
@@ -372,15 +372,49 @@ export class AppServiceDraupnirManager {
     }
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


Gnuxie:
CI failure for appservice just needs to be run again, some problem with setting up synapse

Gnuxie (approved):

PR #575, opened 2024-09-20
Update the README now that we're moving.

Gnuxie:
[Rendered](https://github.com/the-draupnir-project/Draupnir/blob/gnuxie/readme/README.md)


FSG-Cat (approved):
Reads fine to me tho we should have renamed to Neighnir /s

PR #588, opened 2024-10-01
Safe Mode account data config recovery.

Gnuxie:


package.json
```diff
@@ -69,8 +69,8 @@
     "js-yaml": "^4.1.0",
     "jsdom": "^24.0.0",
     "matrix-appservice-bridge": "^9.0.1",
-    "matrix-protection-suite": "npm:@gnuxie/matrix-protection-suite@1.4.0",
-    "matrix-protection-suite-for-matrix-bot-sdk": "npm:@gnuxie/matrix-protection-suite-for-matrix-bot-sdk@1.4.0",
+    "matrix-protection-suite": "npm:@gnuxie/matrix-protection-suite@1.5.0",
+    "matrix-protection-suite-for-matrix-bot-sdk": "npm:@gnuxie/matrix-protection-suite-for-matrix-bot-sdk@1.5.0",
```
Gnuxie, line 73: just gotta bump these to patch the `undefined` value issue then should be good to merge

config/default.yaml
```diff
@@ -211,6 +211,13 @@ protections:
 roomStateBackingStore:
   enabled: false
 
+#safeMode:
+#  # The option for entering safe mode when Draupnir fails to start up.
+#  # - "RecoveryOnly" will only start the bot in safe mode when there are recovery options available.
+#  # - "Never" will never start the bot in safe mode when Draupnir fails to start normally.
+#  # - "Always" will always start the bot in safe mode when Draupnir fails to start normally.
+#  bootOption: RecoveryOnly
```
Gnuxie, line 219: @FSG-Cat @MTRNord If either of you have opinions about this?

FSG-Cat: This Feline approves

FSG-Cat (approved):
This feline is happy with the safe mode configuration options as it gives users flexibility to support their prefered workflows.

config/default.yaml
MTRNord: Seems good with me. It would just be nice to know which possible recover options one could encounter. (as in documentation). Always nice to know what to expect up front for that.

src/safemode/BootOption.ts
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2024 Gnuxie <Gnuxie@protonmail.com>
+//
+// SPDX-License-Identifier: AFL-3.0
+
+export enum SafeModeBootOption {
```
MTRNord, line 5: Hm wouldnt it be smarter to use something like this instead?

```
export const Snack = {
    Apple: 0,
    Banana: 1,
    Orange: 2,
    Other: 3,
} as const;
```

See also https://stackoverflow.com/a/78060803

MTRNord (approved):

src/safemode/BootOption.ts
Gnuxie: No, because enums that are numbers are meaningless while debugging.

Gnuxie: It would be cool if we could use symbols though or if there was a workaround.

Gnuxie: OH right

MTRNord: Hm actually not sure if it must be numbers with this. In the end it will be transpiled to an object with readonly marked elements. So I would guess that Symbols work here. Dont have a PC to try that with me though

Gnuxie: IDK what would we gain from that?

MTRNord: Cleaner transpiled code. The stackoverflow.com has some examples on which enuk variant transpiles to which thing

Gnuxie: Ah ok, do you have particularly strong opinions about that? I don't really feel a need to change it right now, but i'll have it in my mind to think about.

config/default.yaml
Gnuxie: Oki, i'll update that

src/safemode/BootOption.ts
MTRNord: No strong opinions :)

PR #591, opened 2024-10-03
Rename to Draupnir in Appropriate Places

FSG-Cat:
This PR aims to make it so its harder to find places where the bot calls it self Mjolnir without it affecting code in ways that i cant begin to predict.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ config/*
 !config/default.yaml
 
 # Draupnir test data
-mjolnir-registration.yaml
+draupnir-registration.yaml
 test/harness/mjolnir-data/*
 
 # Misc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - 2024-10-04
 
-### Versions v2.0.0-beta.7 and prior
+### Changed
 
-Entrypoint was renamed from `mjolnir-entrypoint.sh` to `draupnir-entrypoint.sh`
+* `Dockerfile`: entry-point was renamed from `mjolnir-entrypoint.sh` to `draupnir-entrypoint.sh`.
+If you have built a Dockerfile based on ours, you may need to make some changes.
 
-`mjolnir-registration.yaml` was renamed to `draupnir-registration.yaml` this change should not have any end user impact.
+* `Dockerfile`: source code was moved from `/mjolnir` to `/draupnir`. If you have built a custom
+docker image based on our Dockerfile based on ours, you may need to make some changes.
+
+* The appservice registration file generator no longer emits `mjolnir-registration.yaml`
+as it has been renamed to `draupnir-registration.yaml`. This is only a concern if you have
+automated tooling that generates a registration file.
+
+## Versions v2.0.0-beta.7 and prior
 
 Please see [Releases](https://github.com/the-draupnir-project/Draupnir/releases) for more information.
+
+<!-- Remove me later. Currently points to releases as this is a Unreleased change.
+As soon as the release is made change the note to point to the release. -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,22 @@
+<!--
+SPDX-FileCopyrightText: 2024 Gnuxie <Gnuxie@protonmail.com>
+
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Changelog
+
+All notable changes to Draupnir will be kept in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased] - 2024-10-04
+
+### Versions v2.0.0-beta.7 and prior
+
+Entrypoint was renamed from `mjolnir-entrypoint.sh` to `draupnir-entrypoint.sh`
+
+`mjolnir-registration.yaml` was renamed to `draupnir-registration.yaml` this change should not have any end user impact.
+
+Please see [Releases](https://github.com/the-draupnir-project/Draupnir/releases) for more information.
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,9 @@ COPY . /tmp/src
 RUN cd /tmp/src \
     && yarn install --network-timeout 100000 \
     && yarn build \
-    && mv lib/ /mjolnir/ \
+    && mv lib/ /draupnir/ \
     && mv node_modules / \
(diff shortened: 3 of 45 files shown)


src/appservice/cli.ts
```diff
@@ -14,12 +14,12 @@ import { IConfig } from "./config/config";
...
 
 /**
- * This file provides the entrypoint for the appservice mode for mjolnir.
+ * This file provides the entrypoint for the appservice mode for draupnir.
  * A registration file can be generated `ts-node src/appservice/cli.ts -r -u "http://host.docker.internal:9000"`
  * and the appservice can be started with `ts-node src/appservice/cli -p 9000 -c your-confg.yaml`.
  */
 const cli = new Cli({
-  registrationPath: "mjolnir-registration.yaml",
+  registrationPath: "draupnir-registration.yaml",
```
Gnuxie, line 22: This will surely break existing deployments?

Dockerfile
```diff
@@ -21,5 +21,5 @@ ENV NODE_ENV=production
 ENV NODE_CONFIG_DIR=/data/config
 
 CMD ["bot"]
-ENTRYPOINT ["./mjolnir-entrypoint.sh"]
+ENTRYPOINT ["./draupnir-entrypoint.sh"]
```
Gnuxie, line 24: I'd prefer to be conservative about changing this, just in case someone else has built their own image based on our own.

src/appservice/cli.ts
FSG-Cat: it wont affect the playbook at all.

Dockerfile
FSG-Cat: You do make a fair point about this. But i also do think that we should break this at some point as we should remove all references to mjolnir from the codebase that are not historical references.

As in if we refer to mjolnir it should be like for example refering to a Mjolnir issue or that this file integrates work from mjolnir.

Gnuxie: > But i also do think that we should break this at some point as we should remove all references to mjolnir from the codebase that are not historical references.

Why?

src/appservice/cli.ts
Gnuxie: > Like im not sure who we would break by renaming the Reg file considering your explicitly passing the reg file location to the AS when running AS mode

Ok

Dockerfile
FSG-Cat: it gives a certain vibe to some people that we refer to our selfs as mjolnir all over the place. This PR should hopefully reduce the chances of end users seeing us call our selfs mjolnir in logs or via command errors.

Gnuxie:
Willing to give this the all clear, @MTRNord, do you have any thoughts before we merge this?

.gitignore
```diff
@@ -7,7 +7,7 @@ config/*
 !config/default.yaml
 
 # Draupnir test data
-mjolnir-registration.yaml
+draupnir-registration.yaml
```
MTRNord, line 10: (just picking a random line of this change)

Make sure this is in the release notes please as it is potentially a breaking change and would break bots if not updated downstream

MTRNord (approved):
looks good to me

.gitignore
Gnuxie: Would it break bots? As far as we know, the path is never referenced directly or implicitly, you have to explicitly provide the path to the registration file to draupnir in the cli, and this is just to generate -- hmm.

@FSG-Cat does the documentation make reference to `mjolnir-registration.yaml` and ask you to move / copy it?

FSG-Cat: Docs never refers to any file called Draupnir or mjolnir that is pre existing. The file that we talk about in docs is created by Draupnir on demand with a file name of our choosing or the docs are wrong currently. (Remember D4A mdad and K8s both use synthetic Reg files that are completely made up by the maintainers of said deployment methods and not generated by Draupnir)

FSG-Cat: This file as far as i can tell is a file that mxtester needs but well i accounted for that.

The more likely to break file would be our entrypoint file if someone is refering to that for some reason. But thats already accounted for as a change that we need to tell people about.

Gnuxie: @FSG-Cat Could you add a `CHANGELOG.md` from the template https://github.com/Gnuxie/tsconfig/blob/main/templates/CHANGELOG.md

Add details for both the entrypoint change and the generated registration file change.

Then at the bottom:

```
## Versions v2.0.0-beta.7 and prior

Please see  https://github.com/the-draupnir-project/Draupnir/releases
```

I don't know if we will use this but it's going to at least help us keep unreleased changes in check.

FSG-Cat: I think its a great idea.

MTRNord: Turns out I _thought_ it was in https://github.com/the-draupnir-project/Draupnir/blob/MTRNord/draupnir4all/docs/appservice.md but a) it wasnt and b) I already used draupnir there...

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+-->
+
+# Changelog
+
+All notable changes to Draupnir will be kept in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased] - 2024-10-04
```
Gnuxie, line 14: @FSG-Cat  you didn't write anything

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Versions v2.0.0-beta.7 and prior
+
+Entrypoint was renamed from `mjolnir-entrypoint.sh` to `draupnir-entrypoint.sh`
+
+`mjolnir-registration.yaml` was renamed to `draupnir-registration.yaml` this change should not have any end user impact.
```
Gnuxie, line 20: You should have wrote this in the unreleased part

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+
+Please see [Releases](https://github.com/the-draupnir-project/Draupnir/releases) for more information.
```
Gnuxie, line 22:
```suggestion
## [Unreleased] - 2024-10-04

### Changed

* `Dockerfile`: entry-point was renamed from `mjolnir-entrypoint.sh` to `draupnir-entrypoint.sh`. If you have built a Dockerfile based on ours, you may need to make some changes.

* `Dockerfile`: source code was moved from `/mjolnir` to `/draupnir`. If you have built a custom docker image based on our Dockerfile based on ours, you may need to make some changes.

* The appservice registration file generator no longer emits `mjolnir-registration.yaml` as it has been renamed to `draupnir-registration.yaml`. This is only a concern if you have automated tooling that generates a registration file.

## Versions v2.0.0-beta.7 and prior

Please see https://github.com/the-draupnir-project/Draupnir/releases for more information.
```

Gnuxie (approved):

PR #592, opened 2024-10-04
Display persistent configs in safe mode

Gnuxie:
image

Task: https://github.com/the-draupnir-project/planning/issues/29


src/safemode/PersistentConfigEditor.ts
```diff
@@ -0,0 +1,99 @@
...
+        )
+      ),
+      new StandardPersistentConfigData(
+        MjolnirEnabledProtectionsDescription as unknown as ConfigDescription<TObject>,
+        new BotSDKAccountDataConfigBackend(
+          client,
+          MjolnirEnabledProtectionsEventType
+        )
+      ),
+    ];
```
Gnuxie, line 67: We really need to give removing the specificity from the mirror a try both here and in MPS.

And remove `as unknown as ConfigDescription` and `as never` that we used to make this work locally for prototyping.

Issue #300, opened 2024-03-20
Running `docker build` without supplying `version.txt` in the repo will fail because `git` isnt in the image


ShadowJonathan:
I encountered this while doing some development and building the image on my VM; Doing a bare clone and then running `docker build` immediately will fail with something like this;

```
root@mjolnir:~# docker-compose build mjol
Building mjol
Step 1/8 : FROM node:20-slim
 ---> 8c3d9f745295
Step 2/8 : COPY . /tmp/src
 ---> c17271863704
Step 3/8 : RUN cd /tmp/src     && yarn install --network-timeout 100000     && yarn build     && mv lib/ /mjolnir/     && mv node_modules /     && mv mjolnir-entrypoint.sh /     && mv package.json /     && mv version.txt /     && cd /     && rm -rf /tmp/*
 ---> Running in 259083d3408e
yarn install v1.22.19
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
Done in 9.73s.
yarn run v1.22.19
$ tsc
$ yarn remove-tests-from-lib && yarn describe-version
$ rm -rf lib/test/ && cp -r lib/src/* lib/ && rm -rf lib/src/
$ (git describe > version.txt.tmp && mv version.txt.tmp version.txt) || true && rm -f version.txt.tmp
/bin/sh: 1: git: not found
Done in 7.71s.
mv: cannot stat 'version.txt': No such file or directory
ERROR: Service 'mjol' failed to build: The command '/bin/sh -c cd /tmp/src     && yarn install --network-timeout 100000     && yarn build     && mv lib/ /mjolnir/     && mv node_modules /     && mv mjolnir-entrypoint.sh /     && mv package.json /     && mv version.txt /     && cd /     && rm -rf /tmp/*' returned a non-zero code: 1
root@mjolnir:~#
```

This isn't noticed in the [Github Actions Flow](https://github.com/the-draupnir-project/Draupnir/blob/main/.github/workflows/docker-hub-release.yml) because there the file is always given, and the step in `yarn build` where it attempts to get it is skipped.

Gnuxie:
This has always been known, but clearly not documented anywhere. You can't just build the docker image without providing `version.txt`. And honestly I didn't know this was something that people were trying to do?

Gnuxie:
[...]

PR #595, opened 2024-10-04
Use multi-stage build in Dockerfile

Gnuxie:
Fixes https://github.com/the-draupnir-project/Draupnir/issues/300

I also tried to use alpine again but rust-crypto is still fucky

with `gcompat`:
```
/node_modules/@matrix-org/matrix-sdk-crypto-nodejs/index.js:250
    throw loadError
    ^

Error: Error relocating /node_modules/@matrix-org/matrix-sdk-crypto-nodejs/matrix-sdk-crypto.linux-x64-musl.node: __register_atfork: symbol not found
    at Module._extensions..node (node:internal/modules/cjs/loader:1586:18)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/node_modules/@matrix-org/matrix-sdk-crypto-nodejs/index.js:175:31)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12) {
  code: 'ERR_DLOPEN_FAILED'
}

```

without `gcompat`:

```
/node_modules/@matrix-org/matrix-sdk-crypto-nodejs/index.js:250
    throw loadError
    ^

Error: Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /node_modules/@matrix-org/matrix-sdk-crypto-nodejs/matrix-sdk-crypto.linux-x64-musl.node)
    at Module._extensions..node (node:internal/modules/cjs/loader:1586:18)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at Module.require (node:internal/modules/cjs/loader:1311:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/node_modules/@matrix-org/matrix-sdk-crypto-nodejs/index.js:175:31)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12) {
  code: 'ERR_DLOPEN_FAILED'
}
```

Testing was done locally by using `docker run --rm -it --network=host -v "/home/user/experiments/Draupnir/test/harness/mjolnir-data:/data" --name=draupnir draupnir-multi-stage bot --draupnir-config /data/config/production.yaml`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN cd /tmp/src \
     && yarn install --frozen-lockfile --production --network-timeout 100000
 
 FROM node:20-slim as final-stage
-COPY --from=git-stamp-stage /tmp/src/version.txt version.txt
+COPY --from=build-stage /tmp/src/version.txt version.txt
 COPY --from=build-stage /tmp/src/lib/ /draupnir/
 COPY --from=build-stage /tmp/src/node_modules /node_modules
 COPY --from=build-stage /tmp/src/draupnir-entrypoint.sh /
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,13 +3,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0 AND AFL-3.0
 
-FROM alpine/git:latest as git-stamp-stage
-COPY . /tmp/src
-RUN cd /tmp/src && git describe > version.txt.tmp && mv version.txt.tmp version.txt
-
 FROM node:20-slim as build-stage
+RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
 COPY . /tmp/src
-COPY --from=git-stamp-stage /tmp/src/version.txt version.txt
+# describe the version.
+RUN cd /tmp/src && git describe > version.txt.tmp && mv version.txt.tmp version.txt
+# build and install
 RUN cd /tmp/src \
     && yarn install --frozen-lockfile --network-timeout 100000 \
     && yarn build \
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,19 +3,24 @@
 #
 # SPDX-License-Identifier: Apache-2.0 AND AFL-3.0
 
-# We can't use alpine anymore because crypto has rust deps.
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -3,19 +3,24 @@
 #
 # SPDX-License-Identifier: Apache-2.0 AND AFL-3.0
 
-# We can't use alpine anymore because crypto has rust deps.
-FROM node:20-slim
+FROM alpine/git:latest as git-stamp-stage
```
Gnuxie, line 6: is it worth doing this or should i just install git in the base image?

MTRNord: I would argue the download overhead of the extra image is larger than an apt install which can be cached in a layer. The size of the whole image is probably larger than a single layer :thinking:

MTRNord:
Hm I think you can then remove the version steps in the CI workflows right? :thinking:

Dockerfile
Gnuxie: oki

Gnuxie:
> Hm I think you can then remove the version steps in the CI workflows right? 🤔

nah, we don't use the docker image in CI

Gnuxie:
@MTRNord I think we just leave at this for now and improve later, what do you think?

MTRNord:
> @MTRNord I think we just leave at this for now and improve later, what do you think?

sure. Whatever works for you :) making good ~~Draupnir~~Dockerimages is hard anyway.

PR #600, opened 2024-10-06
Remove References to Regex in Config Example as they are false

FSG-Cat:
Regex is not supported by wordlist protection so lets strip the references.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -185,6 +185,9 @@ protections:
     #
     # WordList will ban users who use these words when first joining a room, so take caution when selecting them.
     #
+    # The word list protection does not support regular expressions at this time.
+    # The configuration in the past stated support for Regex erroneously.
+    #
     words:
       - "LoReM"
       - "IpSuM"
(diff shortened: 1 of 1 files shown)


config/default.yaml
```diff
@@ -185,10 +185,6 @@ protections:
     #
     # WordList will ban users who use these words when first joining a room, so take caution when selecting them.
     #
-    # For advanced usage, regex can also be used, see the following links for more information;
-    #  - https://www.digitalocean.com/community/tutorials/an-introduction-to-regular-expressions
-    #  - https://regexr.com/
-    #  - https://regexone.com/
```
Gnuxie, line 191: We should probably explicitly say that they don't support regexs

FSG-Cat: meow thats a sensible change but im not sure how you would like it to be worded.

Gnuxie: The word list protection does not support regular expressions at this time.

FSG-Cat: Wording added.

Gnuxie (approved):

PR #637, opened 2024-12-09
`JoinWaveShortCircuitProtection` improvements

Gnuxie:
I'm using this protection to demonstrate commands for changing protection settings. And also the commands for changing the active capability providers. I've fixed and changed a few things along the way.

- The number of joins calculated by the protection was broken. Rather than using a token bucket algorithm, the protection counted the number of joins in a fixed period and then dumped the entire bucket, starting again from 0. We've opted to go for a "lazy" leaky bucket algorithm for simplicity as a replacement.
- We've demonstrated how protections can add commands to Draupnir. The exact mechanism isn't perfect yet but it works. For the `JoinWave` protection this meant adding a status command that would show any recent joins.
- We've updated interface-manager and mps to make sure `protections config add/set/remove` are working as intended.


Gnuxie:
Oh wait i'm a moron

Gnuxie:
Anyways, so it turns out that i've been trying to fix protections add/remove and fucked it up by not basing on main. I'll sort that out. IN the mean time, it looks like somehow we've changed the signature of `internCommand` subtly in a breaking way. grrrr

Gnuxie:
Closing because is being done in: https://github.com/the-draupnir-project/Draupnir/pull/638

PR #638, opened 2024-12-09
Fix protections config set/add/remove commands.

Gnuxie:
- We've updated interface-manager and mps to make sure protections config add/set/remove are working as intended. There were a couple of bugfixes in both.


Gnuxie:
Whoopsie, we need to intern the presentation type translators.

PR #651, opened 2025-01-06
Prevent Draupnir triggering its own Flooding & WordList protections

### LINKED ISSUE #559: WordListProtection tries to ban itself after seeing the ban reason in the managment room

### LINKED ISSUE #579: WordlistProtection and BasicFloodingProtection do not ignore draupnir itself
I wish I didn't have to make /ignore itself, causing me to end up with unwanted content unredacted on my end :/
I have WordListProtection configured to trust after 0 (never), already had disabled BasicFloodingProtection for this exact kind of getting stuck in a loop, this might affect other protections aswell:

image

nexy7574:
Fixes the issue displayed in #579 and #559
(fixes #579, fixes #559)

This PR simply adds a check to the start of the execution chain for WordList and BasicFlooding that just returns an Ok result when the event sender was the client's user ID.
I just tested this and can confirm that this prevents the reported loops in WordList and BasicFlooding, as intended.

### DIFF
--- a/src/protections/WordList.ts
+++ b/src/protections/WordList.ts
@@ -150,6 +150,12 @@ export class WordListProtection
         event.content["body"];
       const roomID = room.toRoomIDOrAlias();
 
+      // If the sender is draupnir, ignore the message
+      if (event["sender"] === this.draupnir.clientUserID) {
+        log.info(`Ignoring message from self: ${event.event_id}`);
+        return Ok(undefined);
+      }
+
       // Check conditions first
       if (minsBeforeTrusting > 0) {
         const roomEntry = this.justJoined.get(roomID);
--- a/src/protections/WordList.ts
+++ b/src/protections/WordList.ts
@@ -138,6 +138,11 @@ export class WordListProtection
     room: MatrixRoomID,
     event: RoomEvent
   ): Promise<ActionResult<void>> {
+    // If the sender is draupnir, ignore the message
+    if (event["sender"] === this.draupnir.clientUserID) {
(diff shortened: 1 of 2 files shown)


Gnuxie (approved):
Meow, thanks for the changes!

Btw you will need to [sign off](https://the-draupnir-project.github.io/draupnir-documentation/contributing#sign-off) (doc link) the PR by just adding `Signed-off-by:` to the description. You can use whatever name you like to do that.

src/protections/WordList.ts
```diff
@@ -138,6 +138,11 @@ export class WordListProtection
     room: MatrixRoomID,
     event: RoomEvent
   ): Promise<ActionResult<void>> {
+    // If the sender is draupnir, ignore the message
+    if (event["sender"] === this.draupnir.clientUserID) {
+      log.info(`Ignoring message from self: ${event.event_id}`);
```
Gnuxie, line 143: whoop, I got too excited, but it's cool and optimistic merging is good. However, these logs would have been maybe ok if they were `debug` rather than `info` (draupnir could get pretty noisy with these protections since we send a lot of events). Personally though I wouldn't have logged at all but there is merit to logging as it would help someone debugging something in future maybe.

Gnuxie: I can fix this for you if you like?

nexy7574: It'd be faster for you to fix it. I noticed I'd left it as INFO almost immediately after you merged lmao

Gnuxie: okie, i can do that tomorrow ^^

PR #652, opened 2025-01-06
Fix rooms being added as a server ban policy

### LINKED ISSUE #458: Room bans written as Server bans.
As issue description says. Room bans are written as server bans. 

the command !draupnir ban !meowmeow:evil.example.com is supposed to result in a ban of that room ID. 

Not a ban of that as a server. Its invalid and the help command atleast doesnt suggest you can override the type detection to tell it no stupid its a room not a server. 

A complete inability to write one of 3 policy types i would classify as a quite severe bug indeed.

nexy7574:
Fixes #458. Thought this was a fluke at first since I only changed one word, but it appears to work reliably and as intended.

### DIFF
--- a/src/commands/Ban.tsx
+++ b/src/commands/Ban.tsx
@@ -155,7 +155,7 @@ export const DraupnirBanCommand = describeCommand({
         return resolvedRoomReference;
       }
       return await policyListEditor.banEntity(
-        PolicyRuleType.Server,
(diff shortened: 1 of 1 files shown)


Gnuxie (approved):
Thanks, that is what was wrong, you can see there that we were literally adding a server rule when the entity being banned was clearly a room! 😩

PR #653, opened 2025-01-06
Remove `--true` flag for `unban` command

### LINKED ISSUE #648: Remove unban `--true` cos it's terrible and confuses new users, make it the default behaviour.
idk why we didn't do this ages ago but whatever

> What I meant is that running with true should work but return a help message saying it’s no longer required as it’s always going to unban everyone

^ also remembe to do that

nexy7574:
Closes #648 by making `--true` the default behaviour.

### DIFF
--- a/src/commands/Unban.ts
+++ b/src/commands/Unban.ts
@@ -132,13 +132,6 @@ export const DraupnirUnbanCommand = describeCommand({
       },
     }
   ),
-  keywords: {
-    keywordDescriptions: {
-      true: {
-        isFlag: true,
-      },
-    },
-  },
   async executor(
     context: DraupnirUnbanCommandContext,
     _info,
@@ -153,7 +146,6 @@ export const DraupnirUnbanCommand = describeCommand({
       issuerManager,
       clientUserID,
       unlistedUserRedactionQueue,
-      managementRoomOutput,
     } = context;
     const policyRoomReference =
       typeof policyRoomDesignator === "string"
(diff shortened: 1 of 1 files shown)


src/commands/Unban.ts
```diff
@@ -132,13 +132,6 @@ export const DraupnirUnbanCommand = describeCommand({
       },
     }
   ),
-  keywords: {
-    keywordDescriptions: {
-      true: {
-        isFlag: true,
-      },
-    },
-  },
```
Gnuxie, line 141: We should probably keep this, just in case someone does provide it and they get confused why draupnir now spits their command back out (draupnir will give them an error if they provide an unknown option).

Just add a comment above `keywords:` saying

`// This is a legacy option to unban the user from all rooms that we now ignore just so providing the option doesn't cause an error.`

Gnuxie (commented):
Thanks, just one tiny change for this one

src/commands/Unban.ts
```diff
@@ -101,8 +101,7 @@ export type DraupnirUnbanCommandContext = {
 };
 
 export const DraupnirUnbanCommand = describeCommand({
-  summary:
-    "Removes an entity from a policy list. If the entity is a glob, then the flag --true must be provided to unban users matching the glob from all protected rooms.",
+  summary: "Removes an entity from a policy list.",
```
Gnuxie, line 104: Ah, now here this is problematic, we've not realised that `--true` had dual purposes:

* The first is wholly terrible, which was to confirm that you wanted to unban users from all rooms
* The second isn't so bad, which is to make sure you understand you are using a glob. Similar to the kick command's `glob` option.

We probably do want that `glob` option to be there and we can pretty much copy what the kick command does there. What do you think?

Gnuxie: Ok nah, that's wrong. Because the whole point of accepting globs is to remove policy rules, not unban users matching globs.

Gnuxie: The message that the command sends to the management room afterwards unbans users matching a glob only because that's what it needs to do when a glob policy rule is removed

PR #654, opened 2025-01-06
Only kick members whose membership is `join`

### LINKED ISSUE #649: Kick command should check the membership before kicking users.
Context: https://matrix.to/#/%21IaWNErZAgQUhGqJXjX%3Amatrix.org/%24oazKe9-9hk77VE-jukGxt8wWYFrqukDxRKZItIsPaFc?via=matrix.org&via=envs.net&via=tchncs.de

`/kick` is being called on users that might already have left or been banned. Some servers are terrible and will treat a call to `/kick` as an unban so this is terrible.

nexy7574:
Fixes #649. Only members who have the `join` membership state will be kicked, which means that not only will banned members not be implicitly unbanned (bug in ruma, and conduits), but draupnir will also no-longer send no-op leave events.

### DIFF
--- a/src/commands/KickCommand.tsx
+++ b/src/commands/KickCommand.tsx
@@ -157,7 +157,7 @@ export const DraupnirKickCommand = describeCommand({
     const usersToKick: UsersToKick = new Map();
     for (const revision of roomsToKickWithin) {
       for (const member of revision.members()) {
-        if (kickRule.test(member.userID)) {
+        if (kickRule.test(member.userID) && member.membership === "join") {
           addUserToKick(
             usersToKick,
             revision.room.toRoomIDOrAlias(),
--- a/src/commands/KickCommand.tsx
+++ b/src/commands/KickCommand.tsx
@@ -159,7 +159,7 @@ export const DraupnirKickCommand = describeCommand({
       for (const member of revision.members()) {
         if (
           kickRule.test(member.userID) &&
-          ["join", "invite"].includes(member.membership)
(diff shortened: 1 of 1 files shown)


src/commands/KickCommand.tsx
```diff
@@ -157,7 +157,10 @@ export const DraupnirKickCommand = describeCommand({
     const usersToKick: UsersToKick = new Map();
     for (const revision of roomsToKickWithin) {
       for (const member of revision.members()) {
-        if (kickRule.test(member.userID)) {
+        if (
+          kickRule.test(member.userID) &&
+          ["invite", "join", "knock"].includes(member.membership)
+        ) {
```
Gnuxie, line 163:
```suggestion
        switch (member.membership)
          case Membership.Knock:
          case Membership.Join:
          case Membership.Invite: {
            if (kickRule.test(membership.userID)) {
              addUserToKick(
                usersToKick,
                revision.room.toRoomIDOrAlias(),
                member.userID
              );
            }
          }
```

Gnuxie: Something like this might be better meow, `Membership` is an enum from `matrix-protection-suite`, what dod you think?

Gnuxie (commented):
meow!!

src/commands/KickCommand.tsx
nexy7574: Oh it is? my bad, looked like a plain string to me :sweat_smile: 

This is likely more efficient anyway :+1:

PR #665, opened 2025-01-11
Don't send the banned word in the ban reason

nexy7574:
This change makes the ban reason when a user sends a banned word generic, and instead sends the banned word in the management room directly. This prevents the banned word from persisting in the room timeline after redacting the original event.

### DIFF
--- a/src/protections/WordList.ts
+++ b/src/protections/WordList.ts
@@ -26,6 +26,7 @@ import {
   UserConsequences,
   Value,
   describeProtection,
+  isError,
 } from "matrix-protection-suite";
 import { Draupnir } from "../Draupnir";
 import {
@@ -33,6 +34,7 @@ import {
   StringRoomID,
   MatrixRoomID,
 } from "@the-draupnir-project/matrix-basic-types";
+import { resultifyBotSDKRequestError } from "matrix-protection-suite-for-matrix-bot-sdk";
 
 const log = new Logger("WordList");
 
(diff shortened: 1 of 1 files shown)


src/protections/WordList.ts
```diff
@@ -193,12 +193,16 @@ export class WordListProtection
...
+        const reason = `Said a bad word. Moderators, consult the management room for more information.`;
         await this.userConsequences.consequenceForUserInRoom(
           roomID,
           event.sender,
           reason
         );
+        await this.draupnir.client.sendMessage(this.draupnir.managementRoomID, {
+          msgtype: "m.notice",
+          body: `Banned ${event.sender} in ${roomID} for saying '${match[0]}'.`,
+        });
```
Gnuxie, line 205: Could you catch this if it errors?

One way could be like this:

```suggestion
        await this.draupnir.client.sendMessage(this.draupnir.managementRoomID, {
          msgtype: "m.notice",
          body: `Banned ${event.sender} in ${roomID} for saying '${match[0]}'.`,
        }).then((_) => Ok(undefined), resultifyBotSDKRequestError);
```

You could also do something like this if you're feeling extra diligent but i don't think there's a need really

```suggestion
        const messageResult = await this.draupnir.client.sendMessage(this.draupnir.managementRoomID, {
          msgtype: "m.notice",
          body: `Banned ${event.sender} in ${roomID} for saying '${match[0]}'.`,
        }).then((_) => Ok(undefined), resultifyBotSDKRequestError);
        if (isError(messageResult) {
          log.error("Failed to send a message to the management room", messageResult.error)
        }
```

Gnuxie (approved):
Thanks once again nexy

src/protections/WordList.ts
nexy7574: Might implement the latter but modify it to include the banned word in the log too, just so that it doesn't have the possibility of simply being lost.

Gnuxie: Okie, that's a great idea ^^

PR #666, opened 2025-01-12
Add --invite to `unban` command

### LINKED ISSUE #622: Be able unban from all protected rooms when entity not on any of the lists
I had recently situation when:

- 3rd party list by mistake included user
- user was removed minute later 

I wanted to unban in all rooms (and ideally re-invite back in those rooms) I thought something like below could come handy

`!draupnir unban <entity> --true`

nexy7574:
This PR closes #622 by adding a `--invite` flag to the `unban` command.

This still needs a bit of work and does not currently work/pass tests, will complete later.

### DIFF
--- a/src/commands/Unban.ts
+++ b/src/commands/Unban.ts
@@ -67,14 +67,11 @@ async function unbanUserFromRooms(
         continue;
       }
       if (rule.test(member.userID)) {
-        let logMessage = `Unbanning ${member.userID} in ${revision.room.toRoomIDOrAlias()}`;
-        if (invite) {
-          logMessage += ", and re-inviting them.";
-        }
         await managementRoomOutput.logMessage(
           LogLevel.DEBUG,
           "Unban",
-          logMessage,
+          `Unbanning ${member.userID} in ${revision.room.toRoomIDOrAlias()}` +
+            (invite ? ", and re-inviting them." : ""),
           revision.room.toRoomIDOrAlias()
         );
         if (!noop) {
--- a/src/commands/Unban.ts
+++ b/src/commands/Unban.ts
@@ -14,6 +14,7 @@ import {
   PolicyListConfig,
   PolicyRoomManager,
   PolicyRuleType,
+  RoomInviter,
   RoomResolver,
   RoomUnbanner,
   SetRoomMembership,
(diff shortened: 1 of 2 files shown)


src/commands/Unban.ts
```diff
@@ -64,17 +67,33 @@ async function unbanUserFromRooms(
...
+        if (invite) {
+          logMessage += ", and re-inviting them.";
+        }
         await managementRoomOutput.logMessage(
           LogLevel.DEBUG,
           "Unban",
-          `Unbanning ${member.userID} in ${revision.room.toRoomIDOrAlias()}`,
+          logMessage,
           revision.room.toRoomIDOrAlias()
         );
```
Gnuxie, line 79: Do you think that this would be simpler? 

```typescript
await managementRoomOutput.logMessage(
  LogLevel.DEBUG,
  "Unban",
  `Unbanning ${member.userID} in ${revision.room.toRoomIDOrAlias()}` + invite ? ', and re-inviting them." : '',
  revision.room.toRoomIDOrAlias()
);
```

src/commands/Unban.ts
```diff
@@ -64,17 +67,33 @@ async function unbanUserFromRooms(
...
+            member.userID
+          );
+          if (isError(inviteResult)) {
+            await managementRoomOutput.logMessage(
+              LogLevel.WARN,
+              "Unban",
+              `Failed to re-invite ${member.userID} to ${revision.room.toRoomIDOrAlias()}`,
+              revision.room.toRoomIDOrAlias()
+            );
+          }
```
Gnuxie, line 96: So when I ran this locally, for some reason this failed every single time, and the reason was that there was a bug in MPS that banned instead of unbanning, that's been fixed now and so we'll have to rebase....

Gnuxie:
It would be nice if we could rewrite the entire command to preview unbanning, probably by stealing code from the standard capability renderer https://github.com/the-draupnir-project/Draupnir/blob/main/src/capabilities/StandardUserConsequencesRenderer.tsx

And then copying the use of `no-confirm` in https://github.com/the-draupnir-project/Draupnir/blob/2ee9f35876c86b7dee681bce1a810d1b5826c0e0/src/safemode/commands/RecoverCommand.tsx

Probably not for this PR though.

src/commands/Unban.ts
nexy7574: Yeah it would be. I was just trying to stick to a line length limit and figured the string in place was quite long lol. I'll replace it.

Issue #669, opened 2025-01-15
Faster membership check option has been removed


FSG-Cat:
Cat discoverd this while 2.0 work for mdad and confirmed with Gnuxie. So thats fun.

PR #670, opened 2025-01-15
Remove Faster membership checks config entry

FSG-Cat:
Fixes https://github.com/the-draupnir-project/Draupnir/issues/669


Gnuxie:
Ah ok, so the reason why we haven't removed this yet is because stable requires it and people will still be linked to the main branch for the config to use. So this is something to merge immediately before or during 2.0 release

PR #673, opened 2025-01-15
Make Room state backing store be default enabled.

FSG-Cat:


Gnuxie (commented):
Can you also make it disabled in this file https://github.com/the-draupnir-project/Draupnir/blob/main/config/harness.yaml? (For integration testing)

Gnuxie:
Thank you Cat

PR #677, opened 2025-01-15
Stop hanging if appservice crashes at startup.

### LINKED ISSUE #503: Appservice hangs with no clear failure if the admin room is not found
What happens is that in my setup Draupnir spawns too early before Synapse is ready to process the alias call and it fails. 
Draupnir gets stuck here.

If draupnir started 30 seconds later this will succeed.

I only noticed this bug today as i went and health checked my bot.

So ye the bot seems healthy on `Version: v2.0.0-beta.4-25-gd1c5d01` if we dont count this bug.

Gnuxie:
Fixes https://github.com/the-draupnir-project/Draupnir/issues/503.

### DIFF
--- a/src/appservice/AppService.ts
+++ b/src/appservice/AppService.ts
@@ -225,16 +225,22 @@ export class MjolnirAppService {
   ): Promise<MjolnirAppService> {
     Logger.configure(config.logging ?? { console: "debug" });
     const dataStore = new PgDataStore(config.db.connectionString);
-    await dataStore.init();
-    const service = await MjolnirAppService.makeMjolnirAppService(
-      config,
-      dataStore,
-      DefaultEventDecoder,
-      registrationFilePath
-    );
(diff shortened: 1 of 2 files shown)


src/appservice/AppService.ts
```diff
@@ -225,16 +225,22 @@ export class MjolnirAppService {
...
+        dataStore,
+        DefaultEventDecoder,
+        registrationFilePath
+      );
+      // The call to `start` MUST happen last. As it needs the datastore, and the mjolnir manager to be initialized before it can process events from the homeserver.
+      await service.start(port);
+      return service;
+    } catch (e: unknown) {
+      log.error("Failed to start the appservice", e);
+      await dataStore.close();
```
Gnuxie, line 241: this is still kinda shit

Gnuxie: I don't like it, there's a `close` method on the MjolnirAppService that we hsould be calling somewhere but we don't ever

PR #679, opened 2025-01-17
Pantalaimon Damnatio Memoriae Stage 1

FSG-Cat:


config/default.yaml
```diff
@@ -11,12 +10,9 @@ rawHomeserverUrl: "https://matrix.org"
 # which would allow using secret management systems such as systemd's service credentials.
 accessToken: "YOUR_TOKEN_HERE"
 
-# Options related to Pantalaimon (https://github.com/matrix-org/pantalaimon)
+# Options related to having the bot login itself.
```
Gnuxie, line 13:
```suggestion
# Options related to having Draupnir login itself.
```

Gnuxie:
We need verify what happens when we use a password with pantalimon and where does that password go, before making a new option option to pass a password secret without understanding what the behavior is.

i guess write down the steps for what happens with the password, what methods are called and if it is persisted anywhere, then add that to the PR. like
The password is assigned to the config
the password from the config is used to do whatever
the access token is then put here
How does the password get disposed of?

Issue #693, opened 2025-01-19
Homeserver ratelimit responses are not properly respected


nexy7574:
When running as a bot user account (with ratelimits), some actions (such as mass-redactions and unban re-inviting) may fail as the current ratelimit handler does not respect the ratelimit response. As a result, the ratelimiter has a low chance of success, as repeated ratelimited requests to Synapse will increase the ratelimit exponentially, faster than Draupnir's fallback does.

nexy7574:
There are two issues here.

The first issue is about to be fixed in an upcoming pull request, where the rate-limit response is incorrectly handled due to a check for an always undefined property in the response error object.

The second issue is that rate-limits are difficult to predict and categorise, meaning that when a rate-limit is encountered, it has to be waited out before additional actions can be taken (for example, if a redact batch succeeds in rooms A, B, and C, but gets rate-limited in D, rooms E, F, and G have to wait until D's rate-limit expires, even if they would not have been rate-limited themselves)

PR #694, opened 2025-01-19
Fix the individual request rate-limit

nexy7574:
This fixes the first problem raised in #693 by making the retry logic only use the fallback calculation if the server-provided retry-after is missing

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -492,10 +492,18 @@ function patchMatrixClientForRetry() {
         return cb(...result);
       } catch (err) {
         // Need to retry.
-        const retryAfterMs =
-          "retryAfterMs" in err
-            ? err.retryAfterMs
-            : attempt * attempt * REQUEST_RETRY_BASE_DURATION_MS;
+        const retryAfterMs = (() => {
+          if (err instanceof MatrixError && err.retryAfterMs !== undefined) {
+            return err.retryAfterMs;
+          } else {
+            LogService.error(
+              "Draupnir.client",
+              "Unable to extract retry_after_ms from error, using fallback to create retry duration",
+              err
+            );
+            return attempt * attempt * REQUEST_RETRY_BASE_DURATION_MS;
+          }
+        })();
         LogService.debug(
           "Draupnir.client",
           `Waiting ${retryAfterMs}ms before retrying ${params.method} ${params.uri}`
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -492,14 +492,10 @@ function patchMatrixClientForRetry() {
         return cb(...result);
       } catch (err) {
         // Need to retry.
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -492,14 +492,10 @@ function patchMatrixClientForRetry() {
...
-            // Use default value.
-          }
-        }
+        const retryAfterMs =
+          "retryAfterMs" in err
+            ? err.retryAfterMs
+            : attempt * attempt * REQUEST_RETRY_BASE_DURATION_MS;
         LogService.debug(
           "Draupnir.client",
           `Waiting ${retryAfterMs}ms before retrying ${params.method} ${params.uri}`
```
Gnuxie, line 501: I wonder if we should log something if we can't extract `retryAfterMs` from the bot-sdk version of the wrapped error at all?

nexy7574: It's my understanding that if it's not present in the bot SDK, either the homeserver didn't return any ratelimiting info (aside from the 429 itself) at all, or it responded with a malformed `retry-after` header *and* didn't include the deprecated `.retry_after_ms` field in JSON. I don't think we really need to account for such an extreme edge case since I doubt it's remotely likely to happen

src/utils.ts
```diff
@@ -492,14 +492,10 @@ function patchMatrixClientForRetry() {
...
-          try {
-            retryAfterMs = Number.parseInt(err.retry_after_ms, 10);
-          } catch (ex) {
-            // Use default value.
-          }
-        }
+        const retryAfterMs =
+          "retryAfterMs" in err
+            ? err.retryAfterMs
+            : attempt * attempt * REQUEST_RETRY_BASE_DURATION_MS;
```
Gnuxie, line 498: How about this? (suggestion code is valid)

```suggestion
        const retryAfterMs = (() => {
          if (err instanceof MatrixError && err.retryAfterMs !== undefined) {
            return err.retryAfterMs;
          } else {
            LogService.error(
              "Draupnir.client",
              "Unable to extract retry_after_ms from error, using fallback to create retry duration",
              err
            );
            return attempt * attempt * REQUEST_RETRY_BASE_DURATION_MS;
          }
        })();
```

src/utils.ts
Gnuxie: The issue is that we don't know if all of our code correctly converts errors to matrix-bot-sdk's `MatrixError`, because of the hacks that we have to do in this file (the utils file) and in matrix-protection-suite-for-matrix-bot-sdk. So I do worry that our own code can complicate things. Either way i think i have covered my concern here https://github.com/the-draupnir-project/Draupnir/pull/694#discussion_r1921607131 :smile_cat:

nexy7574: okay yeah that makes sense, I'll use your suggested changes ^.^

PR #695, opened 2025-01-19
Add an issue template for bug reports.

Gnuxie:


FSG-Cat (approved):
This Seems sensible. Kittys only nitpick is that Room state is default enabled so we could use some non default option. But since your basing this on a 100% real issue its fine.

Issue #703, opened 2025-01-25
Draupnir v2.0.2 can fail to start fetching profile data


Gnuxie:
```
Jan 24 23:50:49 matrix node[3984715]: /home/draupnir/draupnir/node_modules/@gnuxie/typescript-result/dist/Result.js:132
Jan 24 23:50:49 matrix node[3984715]:         return new TypeError(this.toReadableString());
Jan 24 23:50:49 matrix node[3984715]:                ^
Jan 24 23:50:49 matrix node[3984715]: TypeError: ActionException: f734721a-a4fd-42b5-b4a6-85fa2fda076a
Jan 24 23:50:49 matrix node[3984715]: Unable to fetch Draupnir's profile information
Jan 24 23:50:49 matrix node[3984715]: elaborated from: Failed to start Draupnir
Jan 24 23:50:49 matrix node[3984715]: caused by: socket hang up
Jan 24 23:50:49 matrix node[3984715]: from error: Error: socket hang up
Jan 24 23:50:49 matrix node[3984715]: Error: socket hang up
Jan 24 23:50:49 matrix node[3984715]:     at Socket.socketOnEnd (node:_http_client:528:25)
Jan 24 23:50:49 matrix node[3984715]:     at Socket.emit (node:events:530:35)
Jan 24 23:50:49 matrix node[3984715]:     at Socket.emit (node:domain:489:12)
Jan 24 23:50:49 matrix node[3984715]:     at endReadableNT (node:internal/streams/readable:1698:12)
Jan 24 23:50:49 matrix node[3984715]:     at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
Jan 24 23:50:49 matrix node[3984715]:     at ActionException.toExpectError (/home/draupnir/draupnir/node_modules/@gnuxie/typescript-result/dist/Result.js:132:16)
Jan 24 23:50:49 matrix node[3984715]:     at ExpectError (/home/draupnir/draupnir/node_modules/@gnuxie/typescript-result/dist/Result.js:71:28)
Jan 24 23:50:49 matrix node[3984715]:     at Object.expect (/home/draupnir/draupnir/node_modules/@gnuxie/typescript-result/dist/Result.js:76:12)
Jan 24 23:50:49 matrix node[3984715]:     at /home/draupnir/draupnir/lib/index.js:95:68
Jan 24 23:50:49 matrix node[3984715]:     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #714, opened 2025-02-01
Be a little more accepting of servers choking on profile.

Gnuxie:
Fixes https://github.com/the-draupnir-project/Draupnir/issues/703.


Gnuxie:
tbh we should probably check this endpoint does what we think it does??

Gnuxie:
ok so i'm not crazy https://spec.matrix.org/latest/client-server-api/#profiles

Issue #691, opened 2025-01-19
Manual safe mode -> restart causes backing store failure (`SQLITE_CANTOPEN`)


TheArcaneBrony:
```
Sat, 18 Jan 2025 22:12:05 GMT [ERROR] [EventBatch] Caught an exception from the callback for an event batch SqliteError: unable to open database file
    at /usr/share/draupnir/lib/backingstore/better-sqlite3/SqliteRoomStateBackingStore.js:54:34
    at sqliteTransaction (/usr/share/draupnir/node_modules/better-sqlite3/lib/methods/transaction.js:65:24)
    at /usr/share/draupnir/lib/backingstore/better-sqlite3/SqliteRoomStateBackingStore.js:63:13
    at sqliteTransaction (/usr/share/draupnir/node_modules/better-sqlite3/lib/methods/transaction.js:65:24)
    at SqliteRoomStateBackingStore.handleRevision (/usr/share/draupnir/lib/backingstore/better-sqlite3/SqliteRoomStateBackingStore.js:68:13)
    at StandardRoomStateRevisionIssuer.emit (node:events:536:35)
    at StandardRoomStateRevisionIssuer.emit (node:domain:489:12)
    at StandardRoomStateRevisionIssuer.createRevisionFromChanges (/usr/share/draupnir/node_modules/matrix-protection-suite/dist/StateTracking/StandardRoomStateRevisionIssuer.js:67:14)
    at StandardRoomStateRevisionIssuer.createBatchedRevision (/usr/share/draupnir/node_modules/matrix-protection-suite/dist/StateTracking/StandardRoomStateRevisionIssuer.js:76:14)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  code: 'SQLITE_CANTOPEN'
}
```

Issue goes away after properly restarting draupnir.

Gnuxie:
Related to https://github.com/the-draupnir-project/Draupnir/issues/687

Gnuxie:
Hmm struggling to reproduce this one by using restart/safe mode commands @TheArcaneBrony , can you think of anything that could be causing it? Did this happen while Draupnir was in safe mode? When in the process did this happen?

TheArcaneBrony:
this happened after normal mode -> safe mode (ie. !draupnir safemode)

Gnuxie:
Was this propagating to the top level and causing a crash or was it just something you noticed in the logs?

Gnuxie:
[...]

PR #717, opened 2025-02-02
Fix lack of error handling in parts of backing store.

Gnuxie:
Errors propagating here could cause the bot to have only partially synced by propagating all the way up to the sync response handling in the bot-sdk.

https://github.com/the-draupnir-project/Draupnir/issues/691.


Gnuxie:
This isn't gonna fix issues but it'll stop propagation. I'm looking into why the store is so unstable.

PR #723, opened 2025-02-03
Show a protection's compatible capability providers in the capability set section of `!draupnir protections show`.

Gnuxie:
Currently looks like this, we just don't have alternative capabilities yet: 
image


Gnuxie:
The UI is kinda shite atm. We need to make it clear what the capability name is, and also which capability provider in the list is active. And then also change the name so that it's clear the list is of active capability providers.

We should then probably update the tutorial for configuring protections.

Issue #718, opened 2025-02-02
Protections and `SetMembershipPolicyRevisionIssuer`are not being informed of `MembershipChange`s in some situations


Gnuxie:
I don't know how this bug is caused, but it's almost like the revision issuer is completely cutoff.

In the Draupnir where this was observed we had the following:

1. Room protected by draupnir, says so in `!draupnir rooms list`
2. spam bot joins
3. spam bot isn't detected by either `MemberBanSynchronisationProtection` or `RedactionSynchronisationProtection`
4. reprotecting the room (by using `rooms remove`, then `rooms add`) causes `MemberBanSynchronisationProtection` to trigger and ban the user, but not `RedactionSynchronisationProtection`. Which implies that `RedactionSynrhronisationProtection` cannot see the membership change. 

This could mean either:
1. The room is omitted from sync
2. The handles for the protections aren't being called
3. The `SetRoomMembershipRevisionIssuer` is not being issued properly
4. Either the `RoomStateRevisionIssuer` or `RoomMembershipRevisionIssuer`'s are not issuing revisions properly.

Gnuxie:
To aid debugging this, if we don't think this is some weird knock on effect of the backing store breaking. We could add something to the rooms list command that shows how recently the revisions were created.

Gnuxie:
We could also add a `!draupnir rooms state <state type> <key>` command to also help debug this. alongside https://github.com/the-draupnir-project/Draupnir/issues/716

PR #725, opened 2025-02-10
Show the time of the last `RoomStateRevision` for each room in the `!draupnir rooms` command. 

Gnuxie:
We also do this for policy lists.

https://github.com/the-draupnir-project/Draupnir/issues/718

### DIFF
--- a/src/Draupnir.ts
+++ b/src/Draupnir.ts
@@ -298,17 +298,24 @@ export class Draupnir implements Client, MatrixAdaptorContext {
    * This means this is only used in the index.ts.
    */
   public async startupComplete(): Promise<void> {
-    const statusInfo = await draupnirStatusInfo(this);
     try {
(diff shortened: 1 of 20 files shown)


src/Draupnir.ts
```diff
@@ -298,17 +298,24 @@ export class Draupnir implements Client, MatrixAdaptorContext {
...
         "Startup complete. Now monitoring rooms."
       );
+      const statusInfo = await draupnirStatusInfo(this);
+      if (isError(statusInfo)) {
+        log.error(
+          `Was unable to get the status information for draupnir`,
+          statusInfo.error
+        );
+        return;
+      }
```
Gnuxie, line 314: I don't know if we need to do this anymore? since technically i think we got rid of the thing in `draupnirStatusInfo` that could fail

PR #743, opened 2025-03-03
Test more edge cases in unban command

### LINKED ISSUE #740: New unban command probably doesn't unban as many users as it should
If you try to unban a user, say `@alice:example.com` and there is a glob rule for `@*:example.com`. I'm pretty sure the command will only show `@alice:example.com` being unbanned from rooms. Rather than `@bob:example.com` who also matches a glob that is being removed.

### LINKED ISSUE #742: policies aren't removed from unban when there are no matching users in protected rooms
Image

Gnuxie:
Fixes #740 
Fixes #742

### DIFF
--- a/src/commands/unban/Unban.tsx
+++ b/src/commands/unban/Unban.tsx
@@ -11,6 +11,8 @@
 import {
   Ok,
   PolicyRoomManager,
+  PolicyRuleType,
+  Recommendation,
   ResultForUsersInSet,
   RoomInviter,
   RoomResolver,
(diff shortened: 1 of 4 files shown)


Gnuxie:
Ah shit i forgot to test the rendering hang on

Gnuxie:
Honestly it's also kinda sucky, i feel like with a little more work we could get most of the integration test into unit testing and make the API's cleaner. It wouldn't take much work, it's just more work and we've spent ages on this already.

PR #753, opened 2025-03-11
Set up state store if configured in d4all mode

MTRNord:
D4all startup can take up to 120s on my d4all. This will likely help with that :) And it probably wont hurt it.

### DIFF
--- a/src/appservice/AppService.ts
+++ b/src/appservice/AppService.ts
@@ -50,7 +50,6 @@ import {
   isStringRoomID,
 } from "@the-draupnir-project/matrix-basic-types";
 import { SqliteRoomStateBackingStore } from "../backingstore/better-sqlite3/SqliteRoomStateBackingStore";
-import path from "path";
 
 const log = new Logger("AppService");
 /**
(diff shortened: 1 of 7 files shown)


src/appservice/config/config.ts
```diff
@@ -33,6 +33,14 @@ export interface IConfig {
   adminRoom: string;
   /** configuration for matrix-appservice-bridge's Logger */
   logging?: LoggingOpts;
+
+  dataPath: string;
```
Gnuxie, line 37: hmm we'll have to double check this

src/appservice/AppService.ts
```diff
@@ -226,11 +231,22 @@ export class MjolnirAppService {
...
+    const storagePath = path.isAbsolute(config.dataPath)
+      ? config.dataPath
+      : path.join(__dirname, "../", config.dataPath);
+    const eventDecoder = DefaultEventDecoder;
+    const backingStore = config.roomStateBackingStore.enabled
+      ? new SqliteRoomStateBackingStore(
+          path.join(storagePath, "room-state-backing-store.db"),
+          eventDecoder
+        )
+      : undefined;
```
Gnuxie, line 243: Do you think it's worth making the path handling (and creating the store) consistent with src/index.ts somehow? like a utility function in `SqliteRoomStateBackingStore`?

Gnuxie: Pass the event decoder as an argument to this function though

src/appservice/config/config.ts
Gnuxie: ok the usage checks out

Gnuxie, line 37: btw tests are failing because the harness config doesn't include this path

Gnuxie: i think? `src/appservice/config/config.harness.yaml`

Gnuxie: ye

src/appservice/AppService.ts
MTRNord: /me will check/do that yup

src/index.ts
```diff
@@ -46,9 +47,7 @@ void (async function () {
   let bot: DraupnirBotModeToggle | null = null;
   let client: MatrixClient;
   try {
-    const storagePath = path.isAbsolute(config.dataPath)
-      ? config.dataPath
-      : path.join(__dirname, "../", config.dataPath);
+    const storagePath = await realpath(config.dataPath);
```
Gnuxie, line 50: Hmm, how come you use realPath here but then not in `SqliteRoomStateBackingStore.create`?

Gnuxie: Has it changed the behavior?

Gnuxie (approved):

PR #754, opened 2025-03-12
Add schema for the appservice config.

Gnuxie:
We now try to parse the config before destructuring it adhoc in the appservice to help spot errors early.

### DIFF
--- a/src/appservice/config/config.ts
+++ b/src/appservice/config/config.ts
@@ -8,44 +8,109 @@
 // https://github.com/matrix-org/mjolnir
 // </text>
 
+import { Type } from "@sinclair/typebox";
 import * as fs from "fs";
 import { load } from "js-yaml";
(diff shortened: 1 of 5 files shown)


src/appservice/config/config.ts
```diff
@@ -8,44 +8,117 @@
...
+          "Port number for the web API used to access the appservice via the widget",
+      }),
+    },
+    {
+      description:
+        "Config for the web api used to access the appservice via the widget",
+    }
+  ),
+  adminRoom: Type.Union(
+    [StringRoomIDSchema, StringRoomAliasSchema, PermalinkSchema],
```
Gnuxie, line 101: This won't do, we need a `MatrixRoomReference` transform instead of this union because the PermalinkSchema will always fail when a roomID or alias is provided... no i don't know why this isn't failing in integration tests. Needs investigating also.

PR #761, opened 2025-03-14
Support for room policies with hashed entity  and `org.matrix.msc4204.takedown` recommendation in Synapse

Gnuxie:
Story: https://github.com/the-draupnir-project/planning/issues/41
documentation: https://the-draupnir-project.github.io/draupnir-documentation/protections/room-takedown-protection

This PR introduces room takedown functionality into Draupnir.
A new `draupnir takedown` command is added similar to the ban command, but marks entities to be taken down. Because the content is illegal or intolerable. To begin with we only allow takedown of rooms. These takedown policies are sharable with policy lists just like normal bans.

Draupnir responds to takedown policies on Synapse by calling the [room shutdown](https://element-hq.github.io/synapse/latest/admin_api/rooms.html#version-2-new-version) API with the options `block` and `purge`.

The policies that are created by the takedown command are hashed, and this is in order to prevent the room id's being shared directly, and so that we do not create a directory of intolerable content. To be able to use the policies, draupnir therefore needs to be aware of all the rooms that the homeserver is participating in, in order to calculate their hashes and find matching policies, and then takedown the marked rooms.

As part of this process, Draupnir has to "discover" the rooms your server is participating in. This is done via the [synapse-http-antispam](https://the-draupnir-project.github.io/draupnir-documentation/bot/synapse-http-antispam) recently added to draupnir. When draupnir discovers rooms, it will prompt the management room with a notification with some details of the title, room description, and creator. This functionality will be toggleable but will be strongly recommended for servers that have public registration 

Checklist:

- [x] Unit test for room audit log
- [x] Unit test for hash store
- [x] Unit test for room takedown service
- [x] Commands for takedown policies that support only rooms for now
- [x] Consider adding discovered room creator to the hash store so that we can block rooms by creator
       + You wanna do this as an extra table for room details just because of the way the synapse API works
       + Prompt for rooms matching banned creators
       + Consider notifying when a room is discovered
- [x] Integration test for room takedown protection
- [x] Decide if we need to have approval UX (we won't use policies for that yet, just prompts)
- [x] Update MPS dependencies so that CI will pass
- [x] Update documentation for configuring, be sure to talk about the trade offs of enabling `check_event_for_spam`.
- [x] Improve the sqlite queries and add indexes because they're trash at the moment
- [x] Make the sqlite schema transactional and not susceptible to partial upgrades.
- [x] Remove dependence on crypto-js idk how this even hapepned
       - We can do that here by using node:crypto but we can't do it in MPS because of the web situation.
- [x] We're gonna have to include creator with the room_id hash in a discovered room details table, probably with easy access of server name too
- [x] Consider marking rooms below the threshold member count as discovered in memory, but not the db. So that there's chance for them to become discovered when they hit the threshold in the future.
        + We're just going to have to accept there's gaps in the system in terms of room discovery. It's not perfect but getting it right will take a lot more work... One idea is to reuse the approve/disapprove policies to periodically review the public rooms that the server has but the ux for that will need to happen. 
- [x] See if we can tidy up the new stores and combine the databases while we have the time to do so. There will be cross cutting queries between the audit log and the room hash store to be able to find all rooms that haven't been shut down by server or creator
- [x] We will probably need to add server and user hashed policy discovery and integrate that into the same system just so that we can takedown rooms marked by server or user too. We should just be able to pass a policy list revision into the hash store and get all matching rooms from server, user and room rules.
- [x] Think about how approval of policies interacts with the store...
       + They don't get stored. So it's fine.
- [x] Check FIXME's in branch on mps
- [x] Check FIXME's in branch here.
- [x] Make a follow up about approval via using the shutdown command with --no-confirm as a prompt
- [x] Make a follow up about full coverage of room discovery 
       * This is possible by adding an acknowledge room table to the room audit log.
       * We could probably just store "last acknowledged" and offer something to flick through the oldest/unapproved ones. and Add approval seperatley via approval/disapproval policies with different entities.
- [x] Follow up to make discovery log include prompts for OK and Cancel. But also whether to act on bans 
- [x] Follow up about adding user_may_invite check against takedown and banned users / rooms / servers.
- [x] Follow up to Update PolicyListRevision to use a newly created glob trie map for matching globs.
- [x] Follow up for having immediate lookup against hashes in PolicyListRevision
- [x] Follow up for creating a `StringServerName` brand type in matrix-basic-types


Gnuxie:
I need to use in-memory stores within the integration test fixtures

Gnuxie:
We need to update the check_event_for_spam callback to just passthrough all events... since the PDU rate for a typical homeserver is actually really small

PR #766, opened 2025-03-23
Add ban invite retraction protection

TheArcaneBrony:
Standard disclaimer: I did not test the code, as I don't have the required stack available.

The code looks sound to me, but you might have some optimisation ideas.

### DIFF
--- a/src/protections/BanInviteRetraction.ts
+++ b/src/protections/BanInviteRetraction.ts
@@ -1,0 +1,160 @@
+// SPDX-FileCopyrightText: 2025 Emma [it/its] @ Rory& <matrix:u/emma:rory.gay>
+//
+// SPDX-License-Identifier: AFL-3.0
+
+import {
+  AbstractProtection,
+  ActionResult,
+  CapabilitySet,
+  MembershipChange,
+  MembershipChangeType,
+  Ok,
+  PolicyListRevision,
+  PolicyRule,
+  PolicyRuleChange,
+  PolicyRuleType,
+  PowerLevelPermission,
+  ProtectedRoomsSet,
+  Protection,
+  ProtectionDescription,
+  Recommendation,
+  RoomMembershipRevision,
+  SimpleChangeType,
+  Task,
+  describeProtection,
+} from "matrix-protection-suite";
+import { Draupnir } from "../Draupnir";
+import { redactUserMessagesIn } from "../utils";
+import {
+  StringRoomID,
+  StringUserID,
+} from "@the-draupnir-project/matrix-basic-types";
+
+type BanInviteRetractionProtectionDescription = ProtectionDescription<Draupnir>;
+
+export class BanInviteRetractionProtection
+  extends AbstractProtection<BanInviteRetractionProtectionDescription>
+  implements Protection<BanInviteRetractionProtectionDescription> {
+  public constructor(
+    description: BanInviteRetractionProtectionDescription,
+    capabilities: CapabilitySet,
+    protectedRoomsSet: ProtectedRoomsSet,
+    private readonly draupnir: Draupnir,
+  ) {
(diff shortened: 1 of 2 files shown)


src/protections/BanInviteRetraction.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+    }
+    void Task(
+      redactUserMessagesIn(
+        this.draupnir.client,
+        this.draupnir.managementRoomOutput,
+        policy.entity,
+        rooms,
+      ),
+    );
```
Gnuxie, line 73: Why do you want to do this in this protection and not allow redaction synchronisation protection handle it?

src/protections/BanInviteRetraction.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (policy.isGlob()) {
+      this.protectedRoomsSet.allProtectedRooms.forEach((room) =>
+        rooms.push(room.toRoomIDOrAlias()),
+      );
+    } else {
+      for (const roomMembership of this.protectedRoomsSet.setRoomMembership
+        .allRooms) {
+        const membership = roomMembership.membershipForUser(
+          policy.entity as StringUserID,
+        );
```
Gnuxie, line 60: You'd want to guard this with `isStringRoomID` meow

src/protections/BanInviteRetraction.ts
```diff
@@ -0,0 +1,160 @@
...
+    changes: PolicyRuleChange[],
+  ): Promise<ActionResult<void>> {
+    const relevantChanges = changes.filter(
+      (change) =>
+        change.changeType === SimpleChangeType.Added &&
+        change.rule.kind === PolicyRuleType.User,
+    );
+
+    // We don't care about the ramifications of adding a new list here.
+    // The set of changes realistically should be fairly low.
```
Gnuxie, line 87: btw there is a `protectedRoomsSet.setPoliciesMatchingMembership` which can easily fetch you the room members that match policies.

src/protections/BanInviteRetraction.ts
Gnuxie: Redacting the messages won't remove the invites btw? i don't think? if that's the intent

src/protections/BanInviteRetraction.ts
```diff
@@ -0,0 +1,160 @@
...
+          revision.room.toRoomIDOrAlias(),
+          change.userID,
+        ),
+      );
+    }
+    return Ok(undefined);
+  }
+
+  private async retractInvitesFromUserIn(room: StringRoomID, user: StringUserID): Promise<void> {
+    const roomState = await this.draupnir.client.getRoomState(room);
```
Gnuxie, line 132: There's no need to do this. We have access to immediate room state and filters via `this.protectedRoomsSet.setRoomState.getRevisionForRoom`

Gnuxie:
I think generally this should probably be added directly to the redaction synchronisation protection, but we'll continue reviewing it here for now

src/protections/BanInviteRetraction.ts
TheArcaneBrony: Uh, did i miss this? it's supposed to be calling retractPendingInvites or whaterver

TheArcaneBrony:
I'd rather keep it as a separate protection so managers can easily opt in our out of this behavior (though i suppose it could be a config flag...)

src/protections/BanInviteRetraction.ts
TheArcaneBrony: What do you mean?

Gnuxie:
Thanks @TheArcaneBrony , I've used this as the basis for https://github.com/the-draupnir-project/Draupnir/pull/788

PR #796, opened 2025-03-30
Change mentions of Mjölnir to Draupnir

FSG-Cat:
Cat ran into this funny leftover when validating if abuse reporting still works after Cats reworks in mdad. Turns out they do and as it was my first time using these flows Cat discovered this.

### DIFF
--- a/src/report/ReportManager.ts
+++ b/src/report/ReportManager.ts
@@ -96,7 +96,7 @@ enum Kind {
   MODERATION_REQUEST,
   //! An abuse report, as per https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-rooms-roomid-report-eventid
   SERVER_ABUSE_REPORT,
-  //! Mjölnir encountered a problem while attempting to handle a moderation request or abuse report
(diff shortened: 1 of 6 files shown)


src/report/ReportManager.ts
```diff
@@ -590,14 +590,14 @@ interface IUIAction {
...
-   * Detect whether the action may be executed, e.g. whether Mjölnir has
+   * Detect whether the action may be executed, e.g. whether Draupnir has
    * sufficient powerlevel to execute this action.
    *
    * **Security caveat** This assumes that the security policy on whether
    * the operation can be executed is:
    *
    * > *Anyone* in the moderation room and who isn't muted can execute
-   * > an operation iff Mjölnir has the rights to execute it.
+   * > an operation iff Draupnir has the rights to execute it.
```
Mikaela, line 600: This typo could have been fixed

Mikaela (commented):
I was afraid the legacy feature from Mjolnir got "fixed" again, but seems like that didn't happen, while I found a spelling mistake.

src/report/ReportManager.ts
Gnuxie: ono

PR #799, opened 2025-03-31
Homeserver Admin: Restrict resident users automatically based on watched policy rooms

Gnuxie:
We introduce a new concept "restriction" which basically means "can a user send events from this account". Which is important when purging the messages sent from a user while deactivating their account.

We introduce a new protection, the `HomeserverUserPolicyApplication` protection which suspends users automatically when resident users match a policy from watched policy rooms. If the policy contains a reason for `config.automaticallyRedactForReasons`, then we prompt to do a purging deactivate (redact -> deactivate).

Users are also unsuspended as policies are removed.

We also introduce a `--purge-messages` and `--no-confirm` option to the `deactivate` command. This means that the command can now use the user redaction endpoint to redact messages and will shadowban the user while it does this if their account isn't already restricted. The addition of `--no-confirm` means that we now show a preview every time you run the deactivate command where you can review the user details before confirming the deactivation.

  

Follow up 1: https://github.com/the-draupnir-project/Draupnir/issues/813
Follow up 2: https://github.com/the-draupnir-project/Draupnir/issues/814


src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
```diff
@@ -0,0 +1,204 @@
...
+      return false;
+    }
+    if (policy.recommendation === Recommendation.Takedown) {
+      return true;
+    } else if (
+      // FIXME: This would be a policy eligible for purging,
+      // not suspension, any banned local user should be suspended.
+      policy.recommendation === Recommendation.Ban &&
+      this.automaticallyRedactForReasons.some((glob) =>
+        glob.test(policy.reason)
```
Gnuxie, line 57: We need to add some kind of prompt for using the deactivate command with the `purge-messages` option. And then make this suspend in any case.

src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  public handlePolicyChange(changes: PolicyRuleChange[]): void {
+    void Task(
+      (async () => {
+        for (const change of changes) {
+          if (change.changeType === PolicyRuleChangeType.Removed) {
+            continue;
+            // FIXME: Don't we want to unsuspend here?
+            // FIXME: What about modifying from a automaticallyRedactReason to a non redact reason?
```
Gnuxie, line 101: Second FIXME can be ignored.

We do need to unrestrict the user here though

src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
```diff
@@ -0,0 +1,204 @@
...
+    void Task(
+      (async () => {
+        log.debug("Findind local users to suspend at protection enable...");
+        const revision = this.watchedPolicyRooms.currentRevision;
+        // FIXME: We need to also paginate through all supsended, shadowbanned, and locked
+        // users to see if they have a matching policy and recommend that they either
+        // be deactivated or have their restrictions lifted.
+        // probably something for another component of the protection to do.
+        // the users list api lets us also poll by user creation so it's pretty cool.
+        // https://element-hq.github.io/synapse/latest/admin_api/user_admin_api.html#list-accounts-v2
```
Gnuxie, line 151: This will require a follow up issue

src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
```diff
@@ -0,0 +1,204 @@
...
+            );
+          }
+        }
+      })()
+    );
+  }
+
+  public handleProtectionEnable(): void {
+    // FIXME: We probably need to synchronise unrestricted users with the audit log
+    // in the background too.
```
Gnuxie, line 141: This will require a follow up issue

src/protections/HomeserverUserPolicyApplication/HomeserverUserPurgingDeactivate.ts
```diff
@@ -0,0 +1,138 @@
...
+export function isUserAccountRestricted(details: UserDetailsResponse): boolean {
+  return (
+    details.deactivated ||
+    details.shadow_banned ||
+    details.locked ||
+    details.deactivated
+  );
+}
+
+// FIXME: Why isn't glob policy rule allowed here?
```
Gnuxie, line 36: This will require some kind of follow up issue

src/protections/HomeserverUserPolicyApplication/UserRestrictionCapabilityRenderer.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <li>rule: {renderRuleClearText(options.rule)}</li>
+      ) : (
+        <fragment></fragment>
+      )}
+      <li>
+        sender: <code>{options.sender}</code>
+      </li>
+    </ul>
+  );
+}
```
Gnuxie, line 39: Can we get the full policy rendered here please?

image

src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
Gnuxie: we did this

src/protections/HomeserverUserPolicyApplication/UserRestrictionCapabilityRenderer.tsx
Gnuxie: ahh shite will have to be follow up now

src/protections/HomeserverUserPolicyApplication/UserAuditLog.ts
```diff
@@ -0,0 +1,29 @@
+// SPDX-FileCopyrightText: 2025 Gnuxie <Gnuxie@protonmail.com>
+//
+// SPDX-License-Identifier: AFL-3.0
+
+import { Result } from "@gnuxie/typescript-result";
+import { StringUserID } from "@the-draupnir-project/matrix-basic-types";
+import { LiteralPolicyRule } from "matrix-protection-suite";
+import { AccountRestriction } from "matrix-protection-suite-for-matrix-bot-sdk";
+
+export interface UserAuditLog {
```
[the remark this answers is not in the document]
Gnuxie: we did this

src/commands/server-admin/DeactivateCommand.tsx
```diff
@@ -0,0 +1,259 @@
...
+      all their messages?
+      <ul>{renderRuleSummary(policyRule)}</ul>
+    </root>
+  );
+}
+
+// FIXME:
+// hmm now would be a great time to include that feature where we can
+// run commands. but there probably isn't time for that right now.
+// we will just have to send a message, and then the confirmation prompt.
```
Gnuxie, line 232: Follow up issue here, we need to be able to create a prompt to run a command just like the confirmation prompt, but without adding additional arguments like the `--no-confirm` option.

I think this comment is more specifically asking for a way to embed the result and rendering of a command within another event that also includes the confirmation prompt which is probably just a bit silly.

src/protections/HomeserverUserPolicyApplication/HomeserverUserPolicyApplication.ts
```diff
@@ -0,0 +1,234 @@
...
+            log.error(
+              `Failed to check if a user ${policy.entity} is suspended`,
+              isUserRestricted.error
+            );
+          } else if (isUserRestricted.ok) {
+            continue; // user is already suspended
+          }
+          // FIXME: Takedown policie should really automatically deactivate
+          // after a grace period or something.
+          if (
```
Gnuxie, line 135: Consider if we need to feed takedown policies in automatically

Gnuxie:
Integration test will also have to be a follow up too at this point

PR #807, opened 2025-04-11
Make MessageIsMediaProtection correctly check for noop

### LINKED ISSUE #808: MessageIsMediaProtection checks for noop true instead of false when checking for if noop is NOT active.
Description says it all.

FSG-Cat:
This PR fixes the temporary problem while we can fix the actual problem by reworking this whole idea of noop.

Fixes #808


Gnuxie (approved):
Thank you Cat <3

Gnuxie:
Follow up should be to remove the references to noop in protections and make sure noop is causing protections to be instantiated with simulated capabilities.

PR #823, opened 2025-04-29
WebAPI v2

MTRNord:
This is an ongoing task to modernize and extend the web API.

TODOs:

- [ ] Make the commands accessible via the API
- [ ] Expose Statistics
- [ ] Support non owners to fetch available bots if they are in the management room of a bot
- [x] Provide a bot health endpoint


package.json
```diff
@@ -37,7 +37,6 @@
     "@types/nedb": "^1.8.16",
     "@types/node": "^20.14.11",
     "@types/pg": "^8.6.5",
-    "@types/request": "^2.48.12",
```
Gnuxie, line 40: What's going on here?

MTRNord: From what I am seeing, only `src/appservice/Api.ts`, which got deleted, and `test/appservice/utils/webAPIClient.ts` used this. Same with body-parser. So it isnt needed as a dependency anymore. Idk why we only had the types and not the lib though. That probably already was wrong...

PR #824, opened 2025-04-29
Remove the arbitrary provisioning limit

MTRNord:
From all I can see, there is no actual reason this is limited. The db and code both seem to just support this already. There can be an argument that this might instead however be limited via a config option. Not entirely sure.


MTRNord:
No idea why the diff went crazy. It just removes the if case around it...

Gnuxie (commented):
So what's the behaviour if i spam invite `@draupnir-bot:example.com` because i am impatient?

MTRNord:
> So what's the behaviour if i spam invite `@draupnir-bot:example.com` because i am impatient?

lots of bots... :D

maybe a rate limit is a good catch here :D

Gnuxie:
Yeah maybe just a naive cooldown of 1 minute would be the best way forwards.

Gnuxie:
Do you want me to do that or are you ok?

PR #838, opened 2025-05-12
Ensure we run draupnir not as root in Docker

MTRNord:
THIS IS A POSSIBLY BREAKING CHANGE!

This changes the user in docker to the 1000 UID and 1000 GID. Allowing it to run with less and least permissions.
However this also means that potentially the storage volumes are owned by root and not accessible therefore. Users will have to manually fix this. How to migrate will depend on the env.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,10 @@ RUN cd /tmp/src \
     && yarn install --frozen-lockfile --production --network-timeout 100000
 
 FROM node:20-slim as final-stage
+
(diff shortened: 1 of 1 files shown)


MTRNord:
Cat reports that this is safe in mdad and mdad just works fine.

Gnuxie:
The fact this is possibly breaking is a no go and we can't merge this until we are confident that all deployment scenarios are covered.

PR #887, opened 2025-05-30
changelog infos mention searchand special thanks

Gnuxie:


FSG-Cat (approved):
Kitty Katt Approves this message

PR #899, opened 2025-06-14
Add the ability to customise the primary prefix for commands.

### LINKED ISSUE #900: Unable to customise primary prefix for commands or have it be an empty string.
This value is currently hard coded to ! and this is fixed by #899

FSG-Cat:
Fixes #900

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -137,6 +137,11 @@ commands:
   # will address only my_moderator_bot.
   allowNoPrefix: false
 
+  # Controls what symbol will prefix the secondary prefix that is defined in additionalPrefixes.
+  # If you wish the bot to use / commands instead of ! change the value to / and the bot will start responding to / commands.
+
+  symbolPrefixes: "!"
+
   # Any additional bot prefixes that Draupnir will listen to. i.e. adding `mod` will allow `!mod help`.
   additionalPrefixes:
     - "draupnir"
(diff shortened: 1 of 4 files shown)


config/default.yaml
```diff
@@ -137,6 +137,12 @@ commands:
   # will address only my_moderator_bot.
   allowNoPrefix: false
 
+  # Controls what symbol will prefix the secondary prefix that is defined in additionalPrefixes.
```
Gnuxie, line 140:
```suggestion
  # Controls which symbol will prefix a secondary prefix that is described in additionalPrefixes.
  # For example the `!` in `!draupnir`.
```

src/commands/DraupnirCommandDispatcher.ts
```diff
@@ -38,7 +38,7 @@ export function makeDraupnirCommandNormaliser(
   config: IConfig
 ): CommandNormaliser {
   return makeCommandNormaliser(clientUserID, {
-    symbolPrefixes: ["!"],
+    symbolPrefixes: config.commands.symbolPrefixes,
```
Gnuxie, line 41: Can you also check the safe mode and appservice bot mode that they all share this normaliser?

Gnuxie: > You cant configure any of the AS settings currently.

This is not true meow! you can, it loads this config file too!!

PR #901, opened 2025-06-14
Upgrade CI to publish to ghcr.io

FSG-Cat:
Cat is not sure about this PR but the best way to get more attention to it is to publish it so we can start to improve from here.


Gnuxie (commented):
whoopsie

Gnuxie (approved):
ya this will be good

PR #904, opened 2025-06-16
Synapse admin room based room exploration

Gnuxie:
This removes the need for synapse-http-antispam in order to reveal hashed literals for rooms.


Gnuxie (commented):
Theory: I don't think protections manager is disposing of protections that are disabled via a command, or a draupnir that has been disposed. Particularly those that are using the `handlePolicyChange` hook. We've seen a couple of reports about this too, but i think it's happening with the `SynpaseRoomListRoomExplorer`

PR #918, opened 2025-07-16
Attempt to allow Draupnir to use room V12.

Gnuxie:
- Updated to matrix-basic-types 1.4.0 which changes the regex validating room ids.

- Changed the package override so that all dependencies use matrix-basic-types 1.4.0, including the matrix-protection-suite.

- Removed code that tries to store details about discovered rooms in the room takdedown protection. These were unreliable for so many reasons and also are now broken given the room origin cannot be extracted from the room id. Details for why this is can be found in the reviews of https://github.com/matrix-org/matrix-spec-proposals/pull/4291.

### DIFF
--- a/src/protections/RoomTakedown/DiscoveredRoomStore.ts
+++ b/src/protections/RoomTakedown/DiscoveredRoomStore.ts
@@ -4,8 +4,8 @@
 
 import { isError, Ok, Result } from "@gnuxie/typescript-result";
 import {
-  roomIDServerName,
   StringRoomID,
+  userServerName,
 } from "@the-draupnir-project/matrix-basic-types";
 import {
   Logger,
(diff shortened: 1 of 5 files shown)


src/protections/RoomTakedown/DiscoveredRoomStore.ts
```diff
@@ -76,19 +73,9 @@ export class StandardDiscoveredRoomStore
...
-        log.error(
-          "Error storing room details for a room",
-          roomID,
-          details,
-          storeResult
-        );
-      }
+      // TODO: normally we would store details about the room, such as the creator,
+      // and the server, but currently this is unviable.
+      // https://matrix.org/blog/2025/07/security-predisclosure/
```
Gnuxie, line 78: Ok this grand idea doesn't work, we actually do depend on being able to extract the creator in tests.. even though in reality this code is probably populating the database with bad data....

nexy7574: You could always provide a bogus server name if it's not critical

Gnuxie: Ok so all of these changes were unecessary lemao

PR #930, opened 2025-08-12
Make room discovery a synchronous part of the takedown command.

Gnuxie:
Discovery was happening asynchronously in the takedown command as a background task that was triggered by the command.  When a room is used in the takedown command, we do inform Draupnir's room discovery, but it happens as a background task.  Maybe it should not be a background task though, not to fail the command when the informing fails, but to make sure the command is blocked on all side effects happening and then we don't have to write timeouts in tests, which are probably still flakey.


Gnuxie:
This doesn't fix the flakey test, because the flake is from taking down a user. Not a room. Silly!!!

PR #949, opened 2025-09-25
Add TimelineRedactionQueue.

Gnuxie:
This just batches the redactions going on within a single room to the same `/messages` request. And blocks the next request for that room from happening until the current request has completed.

This is necessary because `/messages` pagination is really heavy on homeservers as it can force them to backfill.

There is also some really strange bug in Synapse that is triggered by Draupnir's behaviour
https://github.com/element-hq/synapse/issues/18829.


src/queues/TimelineRedactionQueue.ts
```diff
@@ -0,0 +1,97 @@
...
+    const paginator = this.roomMessages.toRoomMessagesIterator(roomID, {
+      direction: "backwards",
+      limit: this.limit,
+      ...(isGlobInUsers ? {} : { filter: { senders: userIDs } }),
+    });
+    const eventsToRedact: StringEventID[] = [];
+    const paginationResult = await paginator.forEachItem({
+      forEachItemCB: (event) => {
+        if (
+          !isGlobInUsers ||
```
Gnuxie, line 59: We should probably write a comment about why this is here because it looks scary at a glance.

PR #955, opened 2025-09-29
Prompt to watch policy rooms on replacement

Gnuxie:
Part of https://github.com/the-draupnir-project/planning/issues/48

To Decide with @FSG-Cat :

- Do we care about a political situation where a room admin could go rouge and upgrade the room? Currently we do and so we require manual intervention to follow the replacement room, but is this sensible? 

image


FSG-Cat (approved):
Cat has reviewed design choices with Gnuxie and approves this path forwards as its the more cautious path forward.

PR #1029, opened 2026-03-13
Finish Attested Docker Images Ambitions

FSG-Cat:
The foundations for attested docker images are a leftover from when we actually added sbom. So this PR aims to finish that dream.

Draft as i want some attention on this before i say lets go. I dont have the time to properly battle test this today.


FSG-Cat:
Attested docker images build in the @FSG-Cat fork of the repo.

Issue #659, opened 2025-01-09
Status command for Appservice Mode


FSG-Cat:
Bot mode has a status command displaying useful info so why not give one to AS mode.

Gnuxie:
What information would you like to be there?

FSG-Cat:
Useful information could include Bot Count, Total Rooms that the AS is in, Version String Definetively, Health information for the various bots.

The primary thing Cat was thinking about was just Version String information and a bit of misc like that.

FSG-Cat:
The Version string information part of this FR has been implemented in the Version command for Appservice mode.

PR #1059, opened 2026-03-26
Improve Appservice in various ways

FSG-Cat:
Adds the Version command for AS discussed partially in #659 
Also makes it so we have provision limits that are configurable.

And it fixes the Provision Code to actually work correctly again.


FSG-Cat:
Superseded by #1061

Issue #1023, opened 2026-01-19
deprecate `managementRoom` config option and provide "managed" management room instead


Gnuxie:
Our documentation has quite a complicated and length diatribe about e2ee management room being unsupported. In the advent of https://github.com/the-draupnir-project/Draupnir/issues/1022, this is really quite serious.

Instead Draupnir should warn that e2ee management room is not supported and draupnir may refuse to start in a future release.

Anyone on a normal management room will get migrated to a "managed" management room like the appservice has and the room will become inoperable if e2ee is turned on.

Setting up a draupnir in bot mode will require providing an mxid via a cli option and creating the management room for them. Which avoids all the complicated documentation reading.  

context: https://matrix.to/#/!IaWNErZAgQUhGqJXjX:matrix.org/%24sDqKqPS9R9lJ_un4Via9jSF5eeplLCpV5JAw36b1vDo?via=matrix.org&via=breadpirates.chat&via=envs.net

PR #1070, opened 2026-04-07
Zero Touch Provisioning Support Stage 1

FSG-Cat:
Starts us down the path of fixing https://github.com/the-draupnir-project/Draupnir/issues/1023

This PR gets us Zero Touch Provisioning support for AS mode and Bot mode. If your env can take usr and pswd and turn it into access tokens or however you want to do that then well we dont need to do PSWD Auth at all.

If you dont PSWD auth is useful for this and is probably what i will resort to in mdad for Zero Touch Provisioning.

Draft as this PR is very much not even mx-tested because mx-tester decided to say nope cat.


apps/draupnir/src/managedRoomAccountData.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Type } from "@sinclair/typebox";
+import { EDStatic, StringRoomIDSchema } from "matrix-protection-suite";
+import {
+  BotSDKMatrixAccountData,
+  MatrixSendClient,
+} from "matrix-protection-suite-for-matrix-bot-sdk";
+
+export const MANAGEMENT_ROOM_ACCOUNT_DATA_EVENT_TYPE =
+  "space.draupnir.management.room";
+export const ADMIN_ROOM_ACCOUNT_DATA_EVENT_TYPE = "space.draupnir.admin.room";
```
Gnuxie:
```suggestion
export const MANAGEMENT_ROOM_ACCOUNT_DATA_EVENT_TYPE =
  "space.draupnir.management_room";
export const ADMIN_ROOM_ACCOUNT_DATA_EVENT_TYPE = "space.draupnir.admin_room";
```

config/default.yaml
```diff
@@ -52,6 +52,12 @@ acceptInvitesFromSpace: "!example:example.org"
 # Whether Draupnir should report ignored invites to the management room (if autojoinOnlyIfManager is true).
 recordIgnoredInvites: false
 
+# Enable automatic management-room recovery/creation on first boot.
+managedManagementRoom: false
+
+# The initial manager to invite if the management room has to be created.
+# initialManager: "@admin:example.org"
```
Gnuxie, line 58: We should have a different entrypoint for this entirely and these should be cli arguments

Gnuxie: So apprently we can't do that because of the way mdad works and i misunderstood what this was doing anyways.

apps/draupnir/src/config.ts
```diff
@@ -205,6 +209,11 @@ const defaultConfig: IConfig = {
   acceptInvitesFromSpace: "!noop:example.org",
   autojoinOnlyIfManager: true,
   recordIgnoredInvites: false,
+  // FIXME: We are deleting this because we will just use the lack of presence
+  // of the `managementRoom` property to determine whether to make a room or not
+  // and crash if neither the initialAdmin or the managementRoom properties are set.
+  managedManagementRoom: false,
```
Gnuxie: Ok we need to delete this bullshit

apps/draupnir/src/appservice/config/config.example.yaml
```diff
@@ -13,6 +13,9 @@ db:
 # See docs/access_control.md
 adminRoom: "#draupnir-admin:localhost:9999"
 
+# The initial manager to invite if the admin room has to be created.
+# initialManager: "@admin:example.org"
```
Gnuxie, line 17: Doesn't this want to be the default aftre the release?

Gnuxie: ie swap it being commented with admin room

config/default.yaml
```diff
@@ -52,6 +52,10 @@ acceptInvitesFromSpace: "!example:example.org"
 # Whether Draupnir should report ignored invites to the management room (if autojoinOnlyIfManager is true).
 recordIgnoredInvites: false
 
+# The initial manager to invite if the management room has to be created.
+# Leave this commented out when using a pre-existing management room.
+# initialManager: "@admin:example.org"
```
Gnuxie, line 58: Does the fact that this is commented out want to be swapped after release too?

apps/draupnir/src/appservice/config/config.example.yaml
FSG-Cat: Yes you are correct as ZTD should be the default workflow for all new deployments.

config/default.yaml
FSG-Cat: As i said in the other comment your correct.

apps/draupnir/src/appservice/config/config.example.yaml
Gnuxie: Might want to make a follow up issue for htat mew

apps/draupnir/src/config.ts
```diff
@@ -84,7 +86,8 @@ export interface IConfig {
   /** Draupnir will accept invites from members of this space if `autojoinOnlyIfManager` is false. */
   acceptInvitesFromSpace: string | undefined;
   recordIgnoredInvites: boolean;
-  managementRoom: string;
+  initialManager: StringUserID | undefined;
```
Gnuxie: This has to be `string` because nothing is checking it otherwise and we're lying.

apps/draupnir/test/integration/helloTest.ts
```diff
@@ -38,6 +38,9 @@ describe("Test: !help command", function (this: Mocha.Suite) {
     if (draupnir === undefined) {
       throw new TypeError(`setup code is wrong`);
     }
+    if (this.config.managementRoom === undefined) {
+      throw new TypeError("managementRoom is required for this test");
+    }
```
Gnuxie: This is a mistake. Can you tell me why this is a mistake?

apps/draupnir/test/integration/fixtures.ts
```diff
@@ -81,7 +81,9 @@ export const mochaHooks = {
       this.timeout(30000);
       this.lifetime = new StandardLifetime();
       const config = (this.config = configRead());
-      this.managementRoomAlias = config.managementRoom;
+      if (config.managementRoom !== undefined) {
+        this.managementRoomAlias = config.managementRoom;
+      }
```
Gnuxie: Can you write a comment about what is important here? What assumptions have changed?

apps/draupnir/test/integration/helloTest.ts
Gnuxie: It's also untrue and checking the wrong thing. A management room is required for Draupnir to operate, so how can it be that we are starting a test or even have a draupnir instance without one? Obviously we have one in this test.

apps/draupnir/src/config.ts
FSG-Cat: I was under the impression we used to atleast validate this somewhere but yes this makes sense as a change.

Gnuxie: we do but that "validation" narrows the type from `string` to `StringUserID` for us, that's how we know we've done it.

Gnuxie: For you to understand: Otherwise we neuter the use of our types and we have a situation where really `StringUserID` would mean `string | StringUserID` everywhere and that is awful.

FSG-Cat: So string undefined just like management room? Because Cat assumes we still have to keep the whole undefined like managementRoom has.

Gnuxie: yepp, that's right

FSG-Cat: Cat had to revert the commit implemening this for now as Cat did a trace and couldnt find where we validate except here and it broke the build and Cat wants to prioritise having a non broken build while trying if the rest of the codebase even works as advertised.

Edit: Cat should clarify works as advertised as in is this actually working to the degree where we could merge this once we fix this stuff. Because integration tests are good and all but seeing it actually deploy is also valuable.

FSG-Cat:
My WIP mdad PR for ZTD Stage 1 together with the docker image built from 3509497bdcb610d56cca4da9cf1a402d104257a7 by the CI has been successfully test deployed. It survived a full service restart without missbehavior validating that atleast in addition to the validation that the integration tests did succesfully predict this works.

The rest of this work should now focus on wrapping up code quality concerns that Gnuxie has but atleast this does deploy.

apps/draupnir/src/appservice/config/config.example.yaml
TheArcaneBrony, line 17: Wish this was a list option, particularly to allow both the plural usecase (headmate users) to work, and for deployments targetting a group of moderators/being ran for another community in particular.

Gnuxie: The initialManager is purely a bootstrap option to enable a "configure once" workflow for deploying Draupnir. Further invitations can be sent once the initialManager has joined the room and there is no special privilege afforded to the initial manager and they are not considered outside of creating the management room in any way.

Gnuxie: And it's too complex in a critical section of the startup experience for little to no gain given that to make this an array.

PR #1086, opened 2026-04-15
Make Appservice register and setup basic profile information.

### LINKED ISSUE #1081: Make Appservice Register admin account and setup profile information.
Element loves to complain about AS mode missing profile info as we are abusing the fact that the special user essentially is a ghost in the machine that works because an AS says so without actually having to exist. Atleast thats what has been said to cat.

FSG-Cat:
Basic Profile information being setting a display name if none exists.

Fixes #1081

### DIFF
--- a/.changeset/wise-carpets-push.md
+++ b/.changeset/wise-carpets-push.md
@@ -1,0 +1,5 @@
+---
+"draupnir": patch
+---
+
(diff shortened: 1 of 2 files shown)


Gnuxie (approved):
paired

PR #1099, opened 2026-04-20
Split ProjectionNode deltas into a downstream delta and a node state delta

### LINKED ISSUE #123: [Task] Change the `Projection` interface to produce an internal and external delta on `reduceInputs`, `reduceInitialInputs`, and `reduceRebuild`

Gnuxie:
Previously the projection node delta conflated two responsibilities, one was to update the internal state of the projection node and the other was to produce a delta that downstream could consume.

We didn't really notice this until we tried to make a projection that had unrelated nonsense in the downstream delta. So this PR makes the different parts explicit.

We also [tried to look at a  "single step" reducer model](https://github.com/the-draupnir-project/planning/issues/120#issuecomment-4245959599) that only produced a downstream delta. It looks like the single-step model is quite similar to what is actually implemented and bugs in diff code have to be figured out the same way.

It's probably worth exploring the single-step model now that we have discovered what two-step actually entails. 

Closes https://github.com/the-draupnir-project/planning/issues/123

### DIFF
--- a/apps/draupnir/src/commands/WatchPreview.tsx
+++ b/apps/draupnir/src/commands/WatchPreview.tsx
@@ -38,14 +38,15 @@ function watchDeltaForMemberBanIntents(
     );
   return memberBanIntentProjectionNode.reduceInput(
     setMembershipPoliciesRevisionDelta
-  );
+  ).downstreamDelta;
 }
 
 function watchDeltaForServerBanIntents(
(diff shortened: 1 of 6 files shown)


Gnuxie:
Superseded by https://github.com/the-draupnir-project/Draupnir/pull/1100

PR #1103, opened 2026-04-23
Add command to set display name for main bot in AS mode

### LINKED ISSUE #1084: Appservice Name Customsation Main bot
Tracking issue for subfeature as the main issue talks about this enough.

FSG-Cat:
Fixes: https://github.com/the-draupnir-project/Draupnir/issues/1084

### DIFF
--- a/.changeset/brave-carrots-draw.md
+++ b/.changeset/brave-carrots-draw.md
@@ -1,0 +1,5 @@
+---
+"draupnir": patch
+---
+
+Add set display name command to appservice main bot.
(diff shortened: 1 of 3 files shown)


apps/draupnir/test/appservice/integration/displaynameCommandTest.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    expect(
+      extractDisplayName(
+        (await appservice.bridge
+          .getBot()
+          .getClient()
+          .getUserProfile(appservice.botUserID)) as unknown
+      )
+    ).toBe(expectedDisplayname);
```
Gnuxie: btw Draupnir also exposes the `.sendTextCommand` and even `.sendJSCommand` (or something like that). That you can also use in the other integration test

Gnuxie: Since in that test, you use the legacy thing of checking for the green tick. 

But in any case i don't think stuff like this needs integration tests but we'll leave it.

Gnuxie (approved):

PR #1106, opened 2026-04-27
Split `SemanticType` into two steps, describe and verify.

### LINKED ISSUE #132: [Task] Improve `SemanticType` so semantics can be documented more easily without checks
This has become essential because we're creating a lot of projections and projection nodes that have fiddly logic that needs to be documented well.

Gnuxie:
It is a pain in the ass to write the checks but the information itself is still valuable when refactoring. So I'd rather we still encourage people to write this shit down even if they don't write the checks.

Closes https://github.com/the-draupnir-project/planning/issues/132


Gnuxie:
People being me

PR #1114, opened 2026-05-04
Fix Policy Notification Protection only inviting joined users

### LINKED ISSUE #1113: Policy Change Notification Protection doesnt invite invited users breaking AS mode.
The policy change notification protection only checks for the join membership. For fresh provisioned appservices you wont be joined yet. And for Zero Touch Deployment this will also be true.

This does affect bot mode and AS mode its just the symptoms only show up in AS mode and ZTD Bot mode.

FSG-Cat:
Fixes: #1113

### DIFF
--- a/.changeset/lemon-things-tap.md
+++ b/.changeset/lemon-things-tap.md
@@ -1,0 +1,6 @@
+---
+"draupnir": patch
+---
+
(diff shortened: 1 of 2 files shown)


Gnuxie (approved):
I'd say with `membership` join, not with join membership but whatever.

PR #1115, opened 2026-05-04
Zero Touch Deployment Stage 2 Bot Mode

### LINKED ISSUE #1080: ZTD Add support for fetching secrets automatically to bot mode.
Make bot mode fetch its own access tokens. Its rly as simple as that on a feature description level.

This wont be needed for AS mode as AS mode due to being well an appservice actually had in mdad ZTP for this aspect since day 0. Appservices after all use static secrets that you define together with the server that can be essentially whatever as long as its a normal url safe base64 string or whatever the exact char set limit was.

FSG-Cat:
Fixes: https://github.com/the-draupnir-project/Draupnir/issues/1080

### DIFF
--- a/apps/draupnir/src/config.ts
+++ b/apps/draupnir/src/config.ts
@@ -47,6 +47,9 @@ export function getNonDefaultConfigProperties(
   if ("pantalaimon" in nonDefault && isConfigRecord(nonDefault.pantalaimon)) {
     nonDefault.pantalaimon.password = "REDACTED";
   }
+  if ("zeroTouchDeploymentSelfLogin" in nonDefault && isConfigRecord(nonDefault.zeroTouchDeploymentSelfLogin)) {
+    nonDefault.zeroTouchDeploymentSelfLogin.password = "REDACTED";
+  }
   if (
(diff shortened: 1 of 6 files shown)


apps/draupnir/src/config.ts
```diff
@@ -75,6 +81,11 @@ export interface IConfig {
     username: string;
     password: string;
   };
+  zeroTouchDeploymentSelfLogin: {
+    enabled: boolean;
+    username: string;
+    password: string;
+  };
```
Gnuxie, line 88: In external config can we call this something sensible, like what the fuck is this for? What kind of crazy thing are we working around in MAS with this?

FSG-Cat: Gnuxie on a less alarmist note we are also compatible with all matrix homeservers that are currently actually used by people with this method unlike trying to go MAS native and have to maintain a bunch of bullshit to be compatible with all homeserver projects.

Though mdad does only have support for Turnkey user creation (the dependency we use in draupnir for ztd stage 2 mdad) for Synapse as far as i know.

PR #1120, opened 2026-05-05
Fix Ban command renderer giving useless results.

### LINKED ISSUE #551: Missing pills when being prompted for what policy list to use.
Pills and short codes are completely missing when prompted for what policy list to use. 

And logs only say processing command so they are not saying anything is going wrong.

image

### LINKED ISSUE #730: "Would you like to add the ban to a policy list?" only shows room IDs for some clients
**How is Draupnir setup?**

Draupnir4All by `@MTRNord`

**What were you trying to achieve?**

I banned an user by hand and got the prompt asking me whether I would like to add the banned user to the policy list.

However at least Nheko and ElementX iOS don't seem to make the room IDs pills, so I am never fully sure which policy is which.

**What steps did you take to encounter the bug**

1. Ban user outside of Draupnir
2. Switch to Draupnir control room
3. See a question on adding the user to a policy list with mysterious room IDs.

**What is the bug or what did you expect to happen instead?**

I would have expected to know which room is which maybe in format of brackets after the room ID telling me which the policy is or perhaps the policy room name and room ID in brackets after it.

FSG-Cat:
Fixes: #730
Fixes: https://github.com/the-draupnir-project/Draupnir/issues/551

### DIFF
--- a/apps/draupnir/src/commands/Ban.tsx
+++ b/apps/draupnir/src/commands/Ban.tsx
@@ -19,7 +19,11 @@ import {
   RoomResolver,
   WatchedPolicyRooms,
 } from "matrix-protection-suite";
-import { MatrixRoomReference, MatrixUserID, StringUserID } from "@the-draupnir-project/matrix-basic-types";
+import {
+  MatrixRoomReference,
+  MatrixUserID,
+  StringUserID,
+} from "@the-draupnir-project/matrix-basic-types";
 import {
   BasicInvocationInformation,
   MatrixRoomReferencePresentationSchema,
@@ -84,7 +88,10 @@ export const DraupnirBanCommand = describeCommand({
             .getEditablePolicyRoomIDs(clientUserID, PolicyRuleType.User)
             .map((room) =>
               StringPresentationType.wrap(
-                matrixToPermalinkFromClientUser(room.toRoomIDOrAlias(), clientUserID)
+                matrixToPermalinkFromClientUser(
+                  room.toRoomIDOrAlias(),
+                  clientUserID
+                )
               )
             ),
         });
--- a/apps/draupnir/src/commands/Ban.tsx
+++ b/apps/draupnir/src/commands/Ban.tsx
@@ -1,4 +1,5 @@
 // Copyright 2022 - 2025 Gnuxie <Gnuxie@protonmail.com>
+// SPDX-FileCopyrightText: 2026 Catalan Lover <catalanlover@protonmail.com>
 // Copyright 2019 - 2021 The Matrix.org Foundation C.I.C.
 //
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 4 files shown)


apps/draupnir/src/utils.ts
```diff
@@ -31,6 +33,28 @@ export function htmlEscape(input: string): string {
...
 
+/**
+ * Build a matrix.to permalink for a room id or alias using a single via server.
+ * Example: https://matrix.to/#/%21room%3Aexample.com?via=example.com
+ */
+export function matrixToPermalink(roomIdOrAlias: string, viaServer: string): string {
+  return `https://matrix.to/#/${encodeURIComponent(roomIdOrAlias)}?via=${encodeURIComponent(
+    viaServer
+  )}`;
+}
```
Gnuxie, line 44: You should use the `MatrixRoomID`, `MatrixRoomAlias` or `Permalinks` utilities from `matrix-basic-types` for this.

apps/draupnir/src/utils.ts
```diff
@@ -31,6 +33,28 @@ export function htmlEscape(input: string): string {
...
+ * Build a matrix.to permalink for a room id or alias using the bot's homeserver derived
+ * from the provided client MXID.
+ */
+export function matrixToPermalinkFromClientUser(
+  roomIdOrAlias: string,
+  clientUserID: StringUserID
+): string {
+  const via = userServerName(clientUserID);
+  return matrixToPermalink(roomIdOrAlias, via);
+}
```
Gnuxie, line 56: Similarly there is `MatrixUserID` and `Permalinks`

apps/draupnir/test/unit/utils/matrixToPermalink.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    const room = "!wJbHKdEdDUKQRGGImO:feline.support";
+    const via = "feline.support";
+    const result = matrixToPermalink(room, via);
+    expect(result).toBe(
+      `https://matrix.to/#/${encodeURIComponent(room)}?via=${encodeURIComponent(
+        via
+      )}`
+    );
+  });
+});
```
Gnuxie, line 18: No need for this if you use matrix-basic-types mew

apps/draupnir/src/commands/Ban.tsx
```diff
@@ -85,7 +82,11 @@ export const DraupnirBanCommand = describeCommand({
         return Ok({
           suggestions: policyRoomManager
             .getEditablePolicyRoomIDs(clientUserID, PolicyRuleType.User)
-            .map((room) => MatrixRoomIDPresentationType.wrap(room)),
+            .map((room) =>
+              StringPresentationType.wrap(
+                matrixToPermalinkFromClientUser(room.toRoomIDOrAlias(), clientUserID)
+              )
+            ),
```
Gnuxie, line 89: You could use `room.toPermalink()` i think?

Gnuxie, line 89: I don't know if this is a good idea. Some clients will show this as a pill yes, But the presentation renderer for `MatrixRoomID` should really be fixed if it isn't forming pills that clients like.

FSG-Cat: It isnt forming ANYTHING. its spitting bare room IDs and thats it. This PR improves that by spitting bare matrix.to links something clients atleast handle. Atleast Element Web does.

Gnuxie: > It isnt forming ANYTHING. its spitting bare room IDs and thats it. This PR improves that by spitting bare matrix.to links something clients atleast handle. Atleast Element Web does.

In that case, the problem just is simply that when rendering the suggestions to matrix we use the plain text renderer instead of the `DeadDocument` one

PR #1129, opened 2026-05-09
Add Distribution information to status and version command.

FSG-Cat:
This came up today in a discussion about the various ways Draupnir can be installed and distributed that its probably useful to have this little mechanism to define where your Draupnir build originated.

This is essentially Cat CTRL-C/V on the branch metadata.

### DIFF
--- a/apps/draupnir/package.json
+++ b/apps/draupnir/package.json
@@ -9,9 +9,10 @@
   "private": true,
   "scripts": {
     "build": "tsc --project test/tsconfig.json && npm run build:assets",
-    "build:assets": "npm run describe-version && npm run describe-branch",
+    "build:assets": "npm run describe-version && npm run describe-branch && npm run describe-distribution",
     "describe-version": "(git describe > version.txt.tmp && mv version.txt.tmp version.txt) || true && rm -f version.txt.tmp",
     "describe-branch": "(git rev-parse --abbrev-ref HEAD > branch.txt.tmp && mv branch.txt.tmp branch.txt) || true && rm -f branch.txt.tmp",
+    "describe-distribution": "echo \"${DRAUPNIR_DISTRIBUTION:-source}\" > distribution.txt.tmp && mv distribution.txt.tmp distribution.txt",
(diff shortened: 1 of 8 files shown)


apps/draupnir/package.json
```diff
@@ -9,9 +9,10 @@
   "private": true,
   "scripts": {
     "build": "tsc --project test/tsconfig.json && npm run build:assets",
-    "build:assets": "npm run describe-version && npm run describe-branch",
+    "build:assets": "npm run describe-version && npm run describe-branch && npm run describe-distribution",
     "describe-version": "(git describe > version.txt.tmp && mv version.txt.tmp version.txt) || true && rm -f version.txt.tmp",
     "describe-branch": "(git rev-parse --abbrev-ref HEAD > branch.txt.tmp && mv branch.txt.tmp branch.txt) || true && rm -f branch.txt.tmp",
+    "describe-distribution": "echo \"${DRAUPNIR_DISTRIBUTION:-source}\" > distribution.txt.tmp && mv distribution.txt.tmp distribution.txt",
```
Gnuxie, line 15: what is a "build time" distribution?

FSG-Cat: Gnuxie im confused what your refering to?

And as for what the purpose of this is. Its quite simple its to help us get information about where a draupnir comes from.

How was it built and who is to blame for the environment that builds it. For example is the bot built from source then we can start to question the build environment in ways we cant if the bot came from the official CI.

Its also useful for Nix as well Nix currently appends nix to the version string but its tbh better to have that kind of shit be a dedicated attribute.

This is also useful for the 2 or 3 native packages for Draupnir that exist in the wild. Yes OpenSUSE as far as i know has one.

We have no way of knowing the install method unless we coax that information out of the end users and this data would make it easier for us to define a workflow where the first thing we want is that data if the bot boots far enough to cough it up.

TheArcaneBrony: I wonder if this would be better served by... having a JSON file instead?

PR #1133, opened 2026-05-14
Add Quick Access to images comment to PRs

FSG-Cat:
This is part of a larger push to improve the CI. Its just this happens to be a small change that is able to live and die independently of the rest of the work.


Gnuxie:
Cat adding two other random scripts is not a small change

FSG-Cat:
Heres a PR Cat did in Cats fork just to showcase this working well also to actually validate this works.

https://github.com/FSG-Cat/Draupnir/pull/1

PR #1134, opened 2026-05-14
Pin our CI to Sha Digests instead of moving version tags and add Renovate maintenance for this.

FSG-Cat:
This work pins all our CI to sha digests like modern security practices recommends and then adds renovate maintenance for this.

The new workflow added by this PR is to allow the branch protection rule that makes automerges work well work because our CIs dont spit out approvals that get picked up by Branch Protection status checks.


Gnuxie (approved):
Sure i guess, WCGW

PR #1138, opened 2026-05-17
Split Workflow-Gate into Workflow and Script plus Fix Syntax Error.

FSG-Cat:
Gnuxie was right and Cat was wrong. That is tbh how we can summarise this PR. From the bright side Cat is becomming better at this shit.

### DIFF
--- a/.github/scripts/gate.js
+++ b/.github/scripts/gate.js
@@ -15,33 +15,45 @@ module.exports = async ({ github, context }) => {
 
   // Filter out the gate itself and third-party checks
   const knownWorkflows = [
-    'GHCR - Development Branches', 'Docker Hub - Release', 'Docker Hub - Latest', 'Docker Hub - Develop',
-    'GHCR - Release', 'Tests', 'GHCR - Latest', 'Contribution requirements'
+    "GHCR - Development Branches",
+    "Docker Hub - Release",
+    "Docker Hub - Latest",
+    "Docker Hub - Develop",
+    "GHCR - Release",
+    "Tests",
+    "GHCR - Latest",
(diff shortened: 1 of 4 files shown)


.github/scripts/gate.js
```diff
@@ -0,0 +1,63 @@
+// SPDX-FileCopyrightText: 2026 Catalan Lover <catalanlover@protonmail.com>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @param {Object} args
+ * @param {import('@actions/github').getOctokit} args.github
```
Gnuxie, line 7: wth is this bullshit?

FSG-Cat: Intelisense bullshit. Welcome to that intelisense for JS sucks donkey compared to the experience we are used to in typescript parts of Draupnir. https://code.visualstudio.com/docs/languages/javascript#_jsdoc-support

Gnuxie: Honestly i'd rather just have a reminder to merge this shit than automerge if we are going to pulloute or devDependencies with the types for github actions

FSG-Cat: Gnuxie we have multiple actions that run scripts like this actually.

And as for actions we have today the signoff action is actually like that already but inline. And Gnuxie whats so bad about having 1 Dev dep for GH Actions? And thats just because you wanted types for gh actions.

FSG-Cat:
Drafted due to more flaws discovered. Will undraft once this works as intended.

FSG-Cat:
Not needed anymore so closing this.

PR #1139, opened 2026-05-17
Perfect Attestation work and cleanup missing GHCR Attestation

FSG-Cat:
So the story here is quite simple. GHCR used to have its own Production publish image workflow that had attestation.

That part kinda did not survive when we moved on to Unify. He he so Thats fixed now and oh ye we also provide better providence.

Like we now provide both SBOM like before but also SLSA providence. The TLDR on this one is that SBOM is your BOM and SLSA talks about shit like how it was built. Both are important for the goal of all this attestation work tbh.

Like ofc Perfect Reproducibility would also be nice but thats a future goal and tbh.

Final successful attestation https://github.com/the-draupnir-project/Draupnir/attestations/27831236
SLSA Providence Attestation Docs: https://slsa.dev/spec/v1.0/provenance
SBOM Attestation Documentation by Docker: https://docs.docker.com/build/metadata/attestations/sbom/

### DIFF
--- a/.github/workflows/docker-hub-develop.yml
+++ b/.github/workflows/docker-hub-develop.yml
@@ -71,6 +71,7 @@ jobs:
           # prettier-ignore
           outputs: type=image,name=target,annotation-index.org.opencontainers.image.description=Drapunir is a community management platform for Matrix.
           sbom: true
+          provenance: true
           tags: |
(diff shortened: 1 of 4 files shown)


FSG-Cat:
1 Bug located around a missing perm or some other flaw in the attestation pipeline about Artifact Storage attestation so lets fix that before this can be merged.

PR #1143, opened 2026-05-19
Prepare Tests for Merge Queue Support

FSG-Cat:
This change is needed if we want to enable merge queue on this repo.


Gnuxie:
Why?

Gnuxie (approved):
It enables the workflow to run on the merge queue and it wouldn't without this change

PR #1146, opened 2026-05-20
Alpine Docker Image

FSG-Cat:
Starts work on getting us a Alpine Docker image allowing a far lighter image that has less vulnerability surface.

### DIFF
--- a/apps/draupnir/package.json
+++ b/apps/draupnir/package.json
@@ -64,7 +64,8 @@
   "overrides": {
     "@the-draupnir-project/matrix-basic-types": "$the-draupnir-project/matrix-basic-types",
     "@the-draupnir-project/interface-manager": "$the-draupnir-project/interface-manager",
-    "matrix-protection-suite": "$matrix-protection-suite"
+    "matrix-protection-suite": "$matrix-protection-suite",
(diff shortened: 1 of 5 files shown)


FSG-Cat:
This is blocked on being able to build on linux aarch64 musl and the crypto bindings dont build on aarch64 musl.

apps/draupnir/package.json
```diff
@@ -63,7 +63,9 @@
   },
   "overrides": {
     "@the-draupnir-project/matrix-basic-types": "$the-draupnir-project/matrix-basic-types",
-    "@the-draupnir-project/interface-manager": "$the-draupnir-project/interface-manager"
+    "@the-draupnir-project/interface-manager": "$the-draupnir-project/interface-manager",
+    "matrix-protection-suite": "$matrix-protection-suite",
```
Gnuxie: this override doesn't do anything anymore because we no longer alias mps. It's now `@the-draupnir-project/matrix-protection-suite`. In any case we shouldn't need this mew

FSG-Cat: Cat tried to keep the changes from the bump commit minimal. But that is indeed a good thing to point out that could break shit since this is now rebased on a relatively fresh copy of main.

PR #1149, opened 2026-05-21
Integrate Renovate Update Dev Deps and Fix the resulting Linting Errors.

FSG-Cat:
Its known that this PR does not lint cleanly yet its just Cat is not confident about how to Fix the last linting error so Cat decided time to push this if its time for collaboration to fix that last error.

This PR is derived from a slightly earlier version of https://github.com/the-draupnir-project/Draupnir/pull/664 and a PR like this was required due to the whole fucking slew of linting errors.

### DIFF
--- a/apps/draupnir/src/webapis/SynapseHTTPAntispam/SpamCheckEndpointPluginManager.ts
+++ b/apps/draupnir/src/webapis/SynapseHTTPAntispam/SpamCheckEndpointPluginManager.ts
@@ -36,10 +36,13 @@ export class SpamCheckEndpointPluginManager<CBArguments extends unknown[]> {
     this.nonBlockingHandles.add(handle);
   }
 
-  public unregisterHandle(
-    handle: BlockingCallback<CBArguments> | NonBlockingCallback<CBArguments>
+  public unregisterBlockingHandle(handle: BlockingCallback<CBArguments>): void {
+    this.blockingHandles.delete(handle);
+  }
+
+  public unregisterNonBlockingHandle(
+    handle: NonBlockingCallback<CBArguments>
   ): void {
-    this.blockingHandles.delete(handle as BlockingCallback<CBArguments>);
     this.nonBlockingHandles.delete(handle);
   }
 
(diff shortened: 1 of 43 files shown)


apps/draupnir/src/webapis/SynapseHTTPAntispam/SpamCheckEndpointPluginManager.ts
```diff
@@ -36,11 +36,14 @@ export class SpamCheckEndpointPluginManager<CBArguments extends unknown[]> {
...
+    this.blockingHandles.delete(handle);
+  }
+
+  public unregisterNonBlockingHandle(
+    handle: NonBlockingCallback<CBArguments>
   ): void {
-    this.blockingHandles.delete(handle as BlockingCallback<CBArguments>);
-    this.nonBlockingHandles.delete(handle as NonBlockingCallback<CBArguments>);
+    this.nonBlockingHandles.delete(handle);
   }
```
Gnuxie, line 47: But why?

apps/draupnir/test/integration/commands/redactCommandTest.ts
```diff
@@ -32,221 +32,213 @@ describe("Test: The redaction command", function () {
     this.moderator?.stop();
   });
```
Gnuxie, line 33: What is going on in this file? what did you change and why? It's not really possible to tell from the diff mew

apps/draupnir/test/integration/commands/unbanTest.ts
```diff
@@ -103,129 +103,124 @@ async function createWatchedPolicyRoom(
...
+    await createPoliciesBanningUser(policyRoomEditor, falsePositiveUserID);
+    // wait for policies to be detected.
+    await new Promise((resolve) => setTimeout(resolve, 1000));
 
-      // Now we can use the unban command to test the preview has no effects
-      // So the way this can work is we can send the command, get back the event and just know that we can send 'OK' and 'Cancel' to it later and it'll work.
-      const previewResult = (
-        await draupnir.sendTextCommand(
-          draupnir.clientUserID,
-          `!draupnir unban ${falsePositiveUserID}`
```
Gnuxie, line 140: same in this file, it is impossible to tell what is happening, possibly due to also a prettier version bump mixed in with other changes? I can't tell mew

Gnuxie:
maybe changes to prettier need to be isolated in the renovate config too mew

apps/draupnir/src/webapis/SynapseHTTPAntispam/SpamCheckEndpointPluginManager.ts
FSG-Cat: Rules changed and the original way this file was written violates lint rules. I only have the exact lint error from after the fix was run.

PR #1158, opened 2026-07-07
Fix overly strict client capabilities check

### LINKED ISSUE #1102: `ClientCapabilities` schema is too strict
Some homeservers already don't supply `m.set_avatar_url` and `m.set_displayname` https://spec.matrix.org/latest/client-server-api/#mset_displayname-capability

https://spec.matrix.org/latest/client-server-api/#capabilities-negotiation

Essentially we're expecting these properties to be present when they might not be

Context: https://matrix.to/#/!IaWNErZAgQUhGqJXjX:matrix.org/%24HldPWJceifob0yIyx3fIdi9CqyAczQfEuv3ohYFErwo?via=matrix.org&via=breadpirates.chat&via=tchncs.de

bones-was-here:
These are optional per the spec and `m.room_versions` seems to be the only one Draupnir needs. Removed `m.set_avatar_url` and `m.set_displayname` because they're deprecated as well as unused.

Fixes https://github.com/the-draupnir-project/Draupnir/issues/1102

### DIFF
--- a/packages/matrix-protection-suite/src/Client/ClientCapabilityNegotiation.ts
+++ b/packages/matrix-protection-suite/src/Client/ClientCapabilityNegotiation.ts
@@ -28,56 +28,42 @@ const RoomVersionsCapability = Type.Object(
 export type ClientCapabilities = Static<typeof ClientCapabilities>;
 export const ClientCapabilities = Type.Object(
   {
-    "m.change_password": Type.Object(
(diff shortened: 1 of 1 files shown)


Gnuxie (approved):

Gnuxie:
@FSG-Cat why's this validate-docker-build stuck?

PR #1159, opened 2026-07-07
Fix Docker Workflow Validator only running on main repo activity.

FSG-Cat:
This worked fine when the PR originated in the main repo but did not for external PRs and this workflow is safe for external due to no secrets access.

https://github.com/the-draupnir-project/Draupnir/pull/1158 was affected by this and this PR should fix that problem.


Gnuxie:
@FSG-Cat are you happy with this then

PR #1165, opened 2026-07-21
Add support for stable variant of MSC3784

FSG-Cat:


packages/matrix-protection-suite/src/PolicyList/PolicyListRevision.ts
```diff
@@ -27,8 +27,12 @@ import {
 } from "@the-draupnir-project/matrix-basic-types";
 
 /** MSC3784 support. Please note that policy lists predate room types. So there will be lists in the wild without this type. */
-export const POLICY_ROOM_TYPE = "support.feline.policy.lists.msc.v1";
-export const POLICY_ROOM_TYPE_VARIANTS = [POLICY_ROOM_TYPE];
+export const POLICY_ROOM_TYPE = "m.policy";
+export const POLICY_ROOM_TYPE_UNSTABLE = "support.feline.policy.lists.msc.v1";
```
Gnuxie: delete this mew and just put the literal into the variants

FSG-Cat: Meow? Would you like to type that out as a fully formed suggestion if its as basic as Cat assumes. Also Cat is only continuing your exact pattern meow.

PR #1183, opened 2026-09-16
Seperate `Projection` into distinct reader / writer sides and add description for Projection partitions

Gnuxie:
As part of an incremental exploration in developing Draupnir's experimental deterministic projection system architecture, the mutable side of projections needs to be brought out of the regular interface so that the global orchestrator can run derive the next projection node, and not be clobbered by other code. Additionally partition information about projections and the dataflow needs to be available for the orchestrator to even figure out what reducers to run on which projection nodes. So a projection description metaobject has been added that an orchestrator implementation can use to distinguish projections from each other by accessing the partition keys, which are usually a draupnir instance's matrix user ID, a room ID, or both.

This work was started months ago under https://github.com/the-draupnir-project/planning/issues/148
and was done to progress https://github.com/the-draupnir-project/planning/issues/133. https://github.com/the-draupnir-project/planning/issues/120 is blocked because we need a way to distinguish within the legacy revision issuer system between the revision state being built up to the current state (since in Draupnir it effectively starts from nothing each time to catch up to the state of Matrix at startup) and being caught up with the state of matrix at statup and now deriving the next states beyond that. We decided trying to port more of the dataflow pipeline to DPS would help us figure out how to do that.

My current thoughts are that porting the source of dataflow to DPS and stamping it would help us to do that. Since we can then look at what the stamp is at startup and use that to determine whether we have caught up. In theory, because I'm pretty sure that it is a bitch and complete other question how you figure out whether you have caught up or if reality is just faster now. I'm sure there's a big wig answer to that question and it is not a fundamentally unanswered thing that we have to just bodge or make a subjective determination on haha.

Porting the source stream will help because we have to build out DPS from source-to-sink anyways. But getting https://github.com/the-draupnir-project/planning/issues/120 finished would make a major fundamental improvement to the UX of Draupnir and so i think that it is really valuable to focus on that. Specifically because condition reporting improves the state of alarm fatigue in the Draupnir management room by removing most/all of the current noise.


Gnuxie:
Closes https://github.com/the-draupnir-project/planning/issues/152

